PR #30, opened 2020-09-07
nand-177

bnorris2:
making changes to pass the fhirContext through instead of creating a new one every time


query/src/main/java/com/lantanagroup/nandina/query/fhir/r4/cerner/PatientData.java
```diff
@@ -31,8 +31,9 @@
   protected Bundle procedures;
   protected CodeableConcept primaryDx = null;
 
-  public PatientData(EncounterScoop scoop, Patient pat) {
+  public PatientData(EncounterScoop scoop, Patient pat, FhirContext fhirContext) {
     patient = pat;
+    ctx = fhirContext;
```
seanmcilvenna, line 36: when accessing class variables, I prefer to use `this.ctx = XXX;`
really just a preference, but thought I'd share. don't need to make the change for this PR, but worth a thought in future...

seanmcilvenna (approved):

PR #31, opened 2020-09-10
Connectathon

seanmcilvenna:
Adding queryCriteria to configuration so that admins can indicate additional criteria to be added to FHIR server queries
Updating the SANER queries so that it uses multiple code systems to find data Nandina's interested in, as well as to update the fields' codes based on updates to the SANER IG

Setting field in UI to bind to correct answer

Changing AuditEvent creation to set the type isntead of subtype.
Making "type" based on pre-defined enum of options.
Fixing issue with export attempting to store AuditEvents in the query server, instead of the storage server.
Adding in TODO for filling in the "source" of an audit event.


bnorris2 (approved):
looks good

PR #36, opened 2020-10-21
initial commit of unit tests

bnorris2:
initial commit of unit tests

### DIFF
--- a/query/src/main/java/com/lantanagroup/nandina/query/Helper.java
+++ b/query/src/main/java/com/lantanagroup/nandina/query/Helper.java
@@ -1,0 +1,24 @@
+package com.lantanagroup.nandina.query;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.InputStream;
+
+public class Helper {
+
+    public static String convertResourceFileToJson(String fileName) {
+        String jsonString = null;
+        try(InputStream in=Thread.currentThread().getContextClassLoader().getResourceAsStream(fileName)){
+            //pass InputStream to JSON-Library, e.g. using Jackson
+            ObjectMapper mapper = new ObjectMapper();
+            JsonNode jsonNode = mapper.readValue(in,
+                    JsonNode.class);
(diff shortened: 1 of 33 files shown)


query/src/main/java/com/lantanagroup/nandina/query/Helper.java
```diff
@@ -0,0 +1,23 @@
+package com.lantanagroup.nandina.query;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.io.InputStream;
+
+public class Helper {
```
seanmcilvenna, line 8: Should this be in the src/main/test folder somewhere? Are we ever going to use in this production?

query/src/main/java/com/lantanagroup/nandina/query/fhir/r4/cerner/filter/DeathFilter.java
```diff
@@ -23,10 +23,8 @@ public boolean runFilter(PatientData pd) {
     Patient p = pd.getPatient();
     logger.debug("Checking if " + p.getId() + " died");
     if (p.hasDeceasedDateTimeType()) {
-      DateTimeType dtt =	p.getDeceasedDateTimeType();
-      Calendar blah = dtt.toCalendar();
-      Calendar deadDate = p.getDeceasedDateTimeType().toCalendar();
-      if (DateUtils.isSameDay(deadDate.getTime(), reportDate)) {
+      Calendar deathDate = p.getDeceasedDateTimeType().toCalendar();
+      if (DateUtils.isSameDay(deathDate.getTime(), reportDate)) {
```
seanmcilvenna, line 27: Nice

seanmcilvenna (changes requested):
Please add code-comments to each of the top-level filter classes describing what they do.

query/src/main/java/com/lantanagroup/nandina/query/Helper.java
bnorris2: i wasn't sure if we ever would need it in prod...but for now we don't so i guess I can move it to the test folder. @seanmcilvenna

PR #38, opened 2020-10-27
NAND-190 and NAND-194

seanmcilvenna:
Big changes.
Abstracted out UI so that it can be swapped between PIHC and Pillbox.
Abstracted out sender and downloader so that depending on config, the functionality for the "Save to File" and "Send" buttons can be different.
Currently, the UI only displays data, doesn't allow user to change the data. That may change in the future, but that is how we want it for now. With that said, it still follows the same workflow of sending the data to the browser, then when the user wants to download/send the data, the data is sent back to the server.


query/src/main/java/com/lantanagroup/nandina/query/pihc/fhir/r4/cerner/model/DiagnosisReportModel.java
```diff
@@ -0,0 +1,22 @@
+package com.lantanagroup.nandina.query.pihc.fhir.r4.cerner.model;
+
```
bnorris2, line 2: you can use the '@Getter/@Setter' here and then remove the getters and setters if you want

query/src/main/java/com/lantanagroup/nandina/query/pihc/fhir/r4/cerner/model/LabReportModel.java
```diff
@@ -0,0 +1,40 @@
+package com.lantanagroup.nandina.query.pihc.fhir.r4.cerner.model;
+
```
bnorris2, line 2: same here...maybe use lombok instead

query/src/main/java/com/lantanagroup/nandina/query/pihc/fhir/r4/cerner/model/MedReportModel.java
```diff
@@ -0,0 +1,67 @@
+package com.lantanagroup.nandina.query.pihc.fhir.r4.cerner.model;
+
```
bnorris2, line 2: lombok

query/src/main/java/com/lantanagroup/nandina/query/pihc/fhir/r4/cerner/model/PatientReportModel.java
```diff
@@ -0,0 +1,121 @@
+package com.lantanagroup.nandina.query.pihc.fhir.r4.cerner.model;
+
```
bnorris2, line 2: lombok

bnorris2 (approved):
i made a few minor comments, looks good

PR #41, opened 2020-11-18
NAND-204 

bnorris2:
adding initial changes for measure eval ui.

1. I have a todo in the report-body.component.ts for removing the localhost:8081 (need to ask you about this)
2. If we don't like the drop down we can change to select dropdown...i didn't see that in the bootstrap 4 docs.


web/src/main/angular/app/report-body/report-body.component.ts
```diff
@@ -150,9 +159,16 @@ export class ReportBodyComponent implements OnInit {
...
 
+  selectMeasure(selectedItem: string){
+    this.evaluateMeasureButtonText = selectedItem;
+  }
+
   async ngOnInit() {
     // initialize the response date to today by default.
     this.report.date = getFhirNow();
+    // TODO remove localhost:8081. using for testing only. when I use ./api/report/measures it uses 8080 but our api is on 8081
+    this.measureConfigs = await this.http.get<MeasureConfig[]>( 'http://localhost:8081/api/report/measures').toPromise();
```
seanmcilvenna, line 170: this should go inside `MeasureEvalComponent.ngOnInit()` shouldn't it? This seems very specific to `measure-eval` config.

web/src/main/angular/app/report-body/report-body.component.html
```diff
@@ -13,6 +13,18 @@
                     </div>
                 </div>
             </div>
+            <div ngbDropdown class="col-md-4">
```
seanmcilvenna, line 16: After this, I'm going to create a separate set of components for `CriteriaComponent` that uses the same pattern as `ReportBodyComponent` and loads sub-components depending on the `report` configuration... Don't need to do this as part of this ticket/PR, though.

web/src/main/angular/app/report-body/measure-eval/measure-eval.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
```
seanmcilvenna, line 1: Remove this spec.ts file. It's not useful, and we haven't done any E2E testing in Angular as of yet...

web/src/main/angular/assets/config.json
```diff
@@ -1,4 +1,4 @@
 {
   "apiUrl": "http://localhost:8081",
-  "report": "pillbox"
+  "report": "measure-eval"
```
seanmcilvenna, line 3: This is the default configuration. I think this should be overridden by local.json for `measure-eval`, but leave the default configuration to `pihc`, since the server is configured for `pihc` by default, as well.

seanmcilvenna (changes requested):

web/src/main/angular/app/report-body/report-body.component.html
bnorris2: ok

web/src/main/angular/app/report-body/measure-eval/measure-eval.component.spec.ts
bnorris2: roger that

web/src/main/angular/assets/config.json
bnorris2: ok

PR #42, opened 2020-11-24
NAND-206: Create API methods for receiving CDA and FHIR RR's

seanmcilvenna:


api/src/main/java/com/lantanagroup/nandina/controller/RRController.java
```diff
@@ -0,0 +1,94 @@
...
+import java.net.URISyntaxException;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Reportability Response Controller
+ */
+@RestController
+public class RRController extends BaseController {
+  private FhirContext ctx = FhirContext.forR4();
```
bnorris2, line 31: can we get an existing fhirtContext somehow instead of creating a new one?

bnorris2 (approved):
I approve, although i did make one comment, thanks

PR #43, opened 2020-12-01
NAND-205

bnorris2:
adding changes for NAND-205

### DIFF
--- a/query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/FormQuery.java
+++ b/query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/FormQuery.java
@@ -1,0 +1,38 @@
+package com.lantanagroup.nandina.query.measure.fhir.r4;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.rest.client.api.IGenericClient;
+import com.lantanagroup.nandina.QueryReport;
+import com.lantanagroup.nandina.query.BaseFormQuery;
+import org.hl7.fhir.r4.model.MeasureReport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.time.LocalDate;
+
+public class FormQuery extends BaseFormQuery {
+    private static final Logger logger = LoggerFactory.getLogger(com.lantanagroup.nandina.query.measure.fhir.r4.FormQuery.class);
+
+    @Override
+    public void execute() {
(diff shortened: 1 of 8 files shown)


query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/FormQuery.java
```diff
@@ -0,0 +1,38 @@
...
+
+    @Override
+    public void execute() {
+        FhirContext fhirContext = (FhirContext) this.getContextData("fhirContext");
+        MeasureReport measureReport = null;
+        String measureId = this.getContextData("measureId").toString();
+        QueryReport queryReport = (QueryReport) this.getContextData("report");
+
+
+        String url = "https://cqf-ruler.nandina.org/cqf-ruler-r4/fhir/Measure/" + measureId + "/$evaluate-measure?periodStart=" + queryReport.getDate() + "&periodEnd=" + LocalDate.parse(queryReport.getDate()).plusDays(1).toString();
```
seanmcilvenna, line 24: Eventually this should be in configuration. Doesn't have to be part of this PR... but in the next sprint it probably should be.

query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/PrepareQuery.java
```diff
@@ -0,0 +1,45 @@
...
+        String measureId = this.getCriteria().get("measureId");
+
+        IGenericClient fhirQueryClient = (IGenericClient) this.getContextData("fhirQueryClient");
+        ctx = fhirQueryClient.getFhirContext();
+
+        Measure measure = fhirQueryClient.read().resource(Measure.class).withId(measureId).execute();
+
+        Bundle bundle = getBundleTransaction(measure);
+        IGenericClient client = ctx.newRestfulGenericClient("https://cqm-sandbox.alphora.com/cqf-ruler-r4/fhir/");
+        Bundle resp = client.transaction().withBundle(bundle).execute();
```
seanmcilvenna, line 32: I don't _think_ this has to be in a Bundle... I think it can just be `client.update().withResource(measure)` (or something like that). Fine to leave as-is for now, though. It's not harming anything.

seanmcilvenna (approved):

query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/FormQuery.java
bnorris2: yes, i agree

PR #50, opened 2021-01-15
nand-251 

bnorris2:
nand-251 checking for operational outcomes in the exeption for calls to fhir server

### DIFF
--- a/query/src/main/java/com/lantanagroup/nandina/query/BasePrepareQuery.java
+++ b/query/src/main/java/com/lantanagroup/nandina/query/BasePrepareQuery.java
@@ -1,6 +1,7 @@
 package com.lantanagroup.nandina.query;
 
 import ca.uhn.fhir.rest.client.api.IGenericClient;
+import ca.uhn.fhir.rest.server.exceptions.AuthenticationException;
 import com.lantanagroup.nandina.NandinaConfig;
 import org.hl7.fhir.r4.model.Bundle;
 import org.hl7.fhir.r4.model.Resource;
@@ -75,6 +76,9 @@ public abstract class BasePrepareQuery implements IPrepareQuery {
             Bundle bundle = fhirClient.search().byUrl(query).returnBundle(Bundle.class).execute();
 
             return bundle;
+        } catch (AuthenticationException ae) {
+            logger.error("Unable to perform rawSearch() with query " + query + " response body: " + ae.getResponseBody());
+            ae.printStackTrace();
         } catch (Exception ex) {
             this.logger.error("Could not retrieve data from FHIR server " + fhirClient.getServerBase() + " with " + interceptors + " interceptors for " + this.getClass().getName() + ": " + ex.getMessage(), ex);
         }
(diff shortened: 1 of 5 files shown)


query/src/main/java/com/lantanagroup/nandina/query/measure/fhir/r4/PrepareQuery.java
```diff
@@ -104,39 +105,46 @@ public void execute() throws Exception {
...
+                                                "|" +
+                                                p.getIdentifier().get(0).getValue();
+                                patientIds.add(patientId);
+                            }
                         }
-                    }
+                    });
                 });
-            });
+            } catch (AuthenticationException ae) {
```
seanmcilvenna, line 142: it may not always be an `AuthenticationException`, right? It could also be an `HttpResponseException`, correct?

query/src/main/java/com/lantanagroup/nandina/query/scoop/Scoop.java
```diff
@@ -36,6 +37,9 @@ public Bundle rawSearch(IGenericClient fhirClient, String query) {
 					.execute();
 			logger.info(query + " Found " + retBundle.getEntry().size() + " resources");
 			return retBundle;
+		} catch (AuthenticationException ae) {
```
seanmcilvenna, line 40: Same question as above?

query/src/main/java/com/lantanagroup/nandina/query/BaseQuery.java
```diff
@@ -89,6 +90,9 @@ public Bundle rawSearch(String query) {
 			logger.debug(this.getClass().getName() + " executing query: " + query);
 
 			return bundle;
+		} catch (AuthenticationException ae) {
```
seanmcilvenna, line 93: Same question as below? HttpResponseException?

query/src/main/java/com/lantanagroup/nandina/query/BaseQuery.java
```diff
@@ -89,6 +90,9 @@ public Bundle rawSearch(String query) {
...
+			logger.error("Unable to perform rawSearch() with query " + query + " response body: " + ae.getResponseBody());
```
seanmcilvenna, line 94: `logger.error(ae.getResponseBody())` is fine for now, but we should plan in future to replace this with more complex logic to parse the `OperationOutcome` and print out the messages in the OO rather than the whole thing. Would make the logs a little easier to read.

query/src/main/java/com/lantanagroup/nandina/query/BasePrepareQuery.java
```diff
@@ -75,6 +76,9 @@ public Bundle rawSearch(String query) {
             Bundle bundle = fhirClient.search().byUrl(query).returnBundle(Bundle.class).execute();
 
             return bundle;
+        } catch (AuthenticationException ae) {
```
seanmcilvenna, line 79: Same question as below? HttpResponseException?

seanmcilvenna (changes requested):

PR #60, opened 2021-06-08
NHSNLINK-381: Create DocumentReference for new reports

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/controller/BaseController.java
```diff
@@ -1,14 +1,19 @@
...
+import org.hl7.fhir.r4.model.Resource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.security.core.Authentication;
 
 import javax.servlet.http.HttpServletRequest;
 
 public class BaseController {
+  private static final Logger logger = LoggerFactory.getLogger(PatientIdentifierController.class);
```
seanmcilvenna, line 16: `PatientIdentifierController.class` should be `BaseController.class` here.

api/src/main/java/com/lantanagroup/link/api/controller/BaseController.java
```diff
@@ -22,4 +27,28 @@ protected IGenericClient getFhirStoreClient(Authentication authentication, HttpS
...
+    if (!outcome.getCreated() || outcome.getResource() == null) {
+      logger.error("Failed to store/create FHIR resource");
+    } else {
+      logger.debug("Stored FHIR resource with new ID of " + outcome.getResource().getIdElement().getIdPart());
+    }
+  }
+
+  protected void updateResource(Resource resource, IGenericClient fhirStoreClient) {
+
+    if (resource.getMeta().getVersionId() == null){
```
seanmcilvenna, line 42: Remove unused `if` block, or at least comment it out with a reason why

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -301,17 +301,71 @@ public QueryReport generateReport(@AuthenticationPrincipal LinkCredentials user,
...
       FhirHelper.recordAuditEvent(request, fhirStoreClient,  user.getJwt(), FhirHelper.AuditEventTypes.Generate, "Successfully Generated Report");
+
+      // Save measure report
+      String id = RandomStringUtils.randomAlphanumeric(8);
+      measureReport.setId(id);
+      updateResource(measureReport, fhirStoreClient);
+
+      // Save document reference
+      DocumentReference documentReference = getDocumentReference(user, contextData, id);
+      createResource(documentReference, fhirStoreClient);
```
seanmcilvenna, line 315: I generally prefer `this.createResource()` syntax (explicitly using `this.` when calling class-instance methods). Not required, but please consider.

seanmcilvenna: In this case, the same would apply for `this.createResource()` and `this.getDocumentReference()` and `this.updateMeasure()`

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -301,17 +301,71 @@ public QueryReport generateReport(@AuthenticationPrincipal LinkCredentials user,
...
     } catch (Exception ex) {
       logger.error(String.format("Error generating report: %s", ex.getMessage()), ex);
       throw new HttpResponseException(500, "Please contact system administrator regarding this error");
     }
-
     return report;
   }
 
+  private DocumentReference getDocumentReference (LinkCredentials user, HashMap<String, Object> contextData, String id) {
```
seanmcilvenna, line 323: Consider renaming to `generateDocumentReference()`. I like to use
`getX()` for retrieving from a data store or for a getter
`createX()` to create a new model and persist it
`generateX()` to generate an instance of a model, but not persist it (yet)

seanmcilvenna (changes requested):

api/src/main/java/com/lantanagroup/link/api/controller/BaseController.java
arianamihailescu: Completed

seanmcilvenna (approved):

PR #63, opened 2021-06-22
NHSNLINK-391: UI's ReviewComponent implementation

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -447,12 +452,9 @@ public ReportBundle searchReports (Authentication authentication, HttpServletReq
           if (andCond) {
             url += "&";
           }
-          url += "identifier=" + identifier;
+          url += "identifier=" + identifier.replace("|", "%7C");
```
seanmcilvenna, line 455: In JavaScript, there is `encodeURIComponent()`... It would be good to identify a similar function/utility within NHSNLink to use. Would like a single method that I could call to make a String url-friendly. No need for a change in this PR, but consider for future.

web/src/main/angular/app/review/review.component.html
```diff
@@ -1,17 +1,108 @@
...
+                    <button class="btn btn-outline-primary" id="measures" ngbDropdownToggle>{{filter.measure}}</button>
+                    <div ngbDropdownMenu aria-labelledby="measures">
+                        <div *ngFor="let measure of measures">
+                            <button class="dropdown-item"
+                                    id="measureOption" on-click="selectMeasure(measure.name)">{{measure.name}}
+                            </button>
+                        </div>
+                    </div>
+                </div>
+                <!--                <select id="measure" aria-label="measure" class="form-control" [(ngModel)]="filter.measure" (ngModelChange)="search()">-->
```
seanmcilvenna, line 24: Can remove these comments? Looks like the above was decided, instead of a `<select>`

web/src/main/angular/app/review/review.component.html
```diff
@@ -1,17 +1,108 @@
...
+                <div ngbDropdown>
+                    <button class="btn btn-outline-primary" id="status" ngbDropdownToggle>{{filter.status}}</button>
+                    <div ngbDropdownMenu aria-labelledby="status">
+                        <div *ngFor="let status of statuses">
+                            <button class="dropdown-item" on-click="selectStatus(status)">{{status}} </button>
+                        </div>
+
+                    </div>
+                </div>
+                <!--                <select id="status" aria-label="status" class="form-control" [(ngModel)]="filter.status" (ngModelChange)="search()">-->
```
seanmcilvenna, line 43: Remove?

web/src/main/angular/app/review/review.component.html
```diff
@@ -1,17 +1,108 @@
...
+                </div>
+            </th>
+            <th colspan="2">&nbsp;</th>
+
+        </tr>
+        </thead>
+        <tbody>
+        <tr *ngFor="let report of reports">
+            <td>
+                {{getMeasureName(report.measureIdentifier)}}
```
seanmcilvenna, line 71: Nice. Exactly what I had in mind...

web/src/main/angular/app/review/review.component.html
```diff
@@ -1,17 +1,108 @@
...
+                <button class="btn btn-primary" on-click="displayReport(report.id)">{{getLabel(report.status)}}</button>
+            </td>
+        </tr>
+        </tbody>
+        <tfoot>
+        <tr>
+            <td colspan="7">
+                <ngb-pagination class="d-flex justify-content-center"
+                                [maxSize]="5"
+                                [collectionSize]="reports !== undefined?reports.length:0"
```
seanmcilvenna, line 98: I think this can be simplified to `reports?.length || 0`

seanmcilvenna (changes requested):

web/src/main/angular/app/review/review.component.html
arianamihailescu: Fixed it

seanmcilvenna (approved):

PR #66, opened 2021-06-24
Nhsnlink 382

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -279,8 +279,24 @@ private void queryAndStorePatientData (List<String> patientIdentifiers, IGeneric
...
+    DocumentReference documentReference = null;
+    Bundle bundle = fhirStoreClient
+            .search()
+            .forResource(DocumentReference.class)
+            .where(DocumentReference.IDENTIFIER.exactly().systemAndValues(measureIdentifier.getSystem(), measureIdentifier.getValue()))
+            .and(DocumentReference.PERIOD.afterOrEquals().day(startDate))
+            .and(DocumentReference.PERIOD.beforeOrEquals().day(endDate))
+            .returnBundle(Bundle.class)
+            .execute();
+    if (bundle.getEntry().size() > 0) {
```
seanmcilvenna, line 292: Should maybe be `== 1` and have
`else { throw error }`

web/src/main/angular/app/report-body/report-body.component.ts
```diff
@@ -83,10 +83,17 @@ export class ReportBodyComponent implements OnInit {
...
 
       try {
-        const updatedReport = await this.reportService.generate(this.report);
+        const updatedReport = await this.reportService.generate(this.report, false);
         Object.assign(this.report, updatedReport);
       } catch (ex) {
-        this.toastService.showException('Error generating report', ex);
+        if (ex.error.message.includes("status code: 409")){
+          if (confirm(ex.error.message)) {
+            const updatedReport = await this.reportService.generate(this.report, true);
```
seanmcilvenna, line 91: We should catch exceptions from this, as well, in case there was an error during re-generation and we can display a message to the user.

seanmcilvenna (changes requested):

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
arianamihailescu: Throw  an error if we have more than one documents for same period (it should not happen)? In that case what should we do just display an error that are more than 1 reports for that period and that is all?

web/src/main/angular/app/report-body/report-body.component.ts
arianamihailescu: Added that

PR #70, opened 2021-07-01
NHSNLINK-292

seanmcilvenna:


arianamihailescu (commented):
It looks good. Just one comment: it used to be some code in storeLatestMeasure  that does not exist in storeReportBundleResources  to verify that every entry in  the Measure bundle has all it needs before being stored on the server:  like method:PUT, url.  I guess it is not needed but just wanted to verify.

arianamihailescu (approved):

PR #75, opened 2021-07-06
NHSNLINK-461

seanmcilvenna:


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -100,64 +101,38 @@ private String storeReportBundleResources(Bundle bundle, IGenericClient fhirStor
     return measureId;
   }
 
-  private void resolveMeasure (Map<String, String> criteria, IGenericClient fhirStoreClient, Map<String, Object> contextData) throws Exception {
-    String reportDefId = criteria.get("measureId");
-
+  private void resolveMeasure(ReportCriteria criteria, ReportContext context) throws Exception {
```
seanmcilvenna, line 104: Passing around a `ReportCriteria` and `ReportContext` instance now, that is strongly typed, instead of using a HashMap; a safer approach.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -100,64 +101,38 @@ private String storeReportBundleResources(Bundle bundle, IGenericClient fhirStor
...
-      logger.error("Error storing resources for the report definition " + reportDefId + ": " + ex.getMessage());
+      logger.error("Error storing resources for the report definition " + criteria.getReportDefId() + ": " + ex.getMessage());
       throw new Exception("Error storing resources for the report definition: " + ex.getMessage());
     }
-
-    contextData.put("measureBundle", measureBundle);
-    contextData.put("measureIdentifier", measureBundle.getIdentifier());
   }
 
-  private Map<String, String> getCriteria (HttpServletRequest request, QueryReport report) {
```
seanmcilvenna, line 131: Since "criteria" is now strongly typed, no need to produce a `Map` from the request (and `QueryReport` is no longer passed into the `generateReport` method).

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -269,47 +244,48 @@ private DocumentReference getDocumentReferenceByMeasureAndPeriod (Identifier mea
             .cacheControl(new CacheControlDirective().setNoCache(true))
             .execute();
     int size = bundle.getEntry().size();
-    if (size > 0 ) {
```
seanmcilvenna, line 272: I wonder if we can get IntelliJ to make these coding patterns consistent...
I think you have used:
```
if(...) {
}
else {
}
```

And I have historically used:

```
if (...) {
} else {
}
```

(space after the `if` and `(` and no line-break between the `else` and `{`

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -269,47 +244,48 @@ private DocumentReference getDocumentReferenceByMeasureAndPeriod (Identifier mea
...
         throw new Exception("We have more than 1 report for the selected measure and report date.");
       }
     }
     return documentReference;
   }
 
   @PostMapping("/$generate")
-  public QueryReport generateReport (@AuthenticationPrincipal LinkCredentials user, Authentication authentication, HttpServletRequest request, @RequestBody() QueryReport report, boolean regenerate) throws Exception {
-
+  public GenerateResponse generateReport(
```
seanmcilvenna, line 258: Passing in the criteria as request params, instead of having the browser pass a complex `QueryReport` object that is largely useless.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -269,47 +244,48 @@ private DocumentReference getDocumentReferenceByMeasureAndPeriod (Identifier mea
...
+          @AuthenticationPrincipal LinkCredentials user,
+          Authentication authentication,
+          HttpServletRequest request,
+          @RequestParam("reportDefId") String reportDefId,
+          @RequestParam("periodStart") String periodStart,
```
seanmcilvenna, line 263: `periodStart` and `periodEnd` are now passed forward from the browser separately, in anticipation of the UI (in the future) potentially letting the user select a range of dates, vs. a single date.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -323,16 +299,17 @@ public QueryReport generateReport (@AuthenticationPrincipal LinkCredentials user
       } else {
         id = null != existingDocumentReference ? existingDocumentReference.getMasterIdentifier().getValue() : "";
       }
-      contextData.put("reportId", id);
-      MeasureReport measureReport = MeasureEvaluator.generateMeasureReport(criteria, contextData, this.config, fhirStoreClient);
+      context.setReportId(id);
+      response.setReportId(id);
+      MeasureReport measureReport = MeasureEvaluator.generateMeasureReport(criteria, context, this.config);
```
seanmcilvenna, line 304: `fhirStoreClient` is already in the `context`... Changed `MeasureEvaluator.generateMeasureReport` to just use that instead of passing it in as a separate variable. Should probably do some `if (context == null || context.getFhirStoreClient() == null)` checking within the `MeasureEvaluator.generateMeasureReport` in the future, though.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -347,19 +324,18 @@ public QueryReport generateReport (@AuthenticationPrincipal LinkCredentials user
       logger.error(String.format("Error generating report: %s", ex.getMessage()), ex);
       throw new ResponseStatusException(HttpStatus.INTERNAL_SERVER_ERROR, "Please contact system administrator regarding this error.");
     }
-    return report;
-  }
 
-  private DocumentReference generateDocumentReference (LinkCredentials user, HashMap<String, Object> contextData, String id) {
+    return response;
```
seanmcilvenna, line 328: I originally intended to just return a `String` that represents the report id.. But, ran into serialization issues with Spring, and it didn't treat it as JSON coming out of the REST API, and the UI code was expecting JSON. So, I made it a complex object. Right now, it only includes `reportId`, but in the future I think it may contain more information than that... We'll see.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -371,9 +347,9 @@ private DocumentReference generateDocumentReference (LinkCredentials user, HashM
     CodeableConcept type = new CodeableConcept();
     List<Coding> codings = new ArrayList<>();
     Coding coding = new Coding();
-    coding.setCode("55186-1");
-    coding.setSystem("http://loinc.org");
-    coding.setDisplay("Measure Document");
+    coding.setCode(Constants.DocRefCode);
```
seanmcilvenna, line 350: Trying to move a lot of static strings to `Constants` so that we only have a single place to look for most of the static strings.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -404,7 +379,7 @@ private DocumentReference generateDocumentReference (LinkCredentials user, HashM
    * @throws Exception Thrown when the configured sender class is not found or fails to initialize
    */
   @PostMapping("/send")
-  public void send (@RequestBody() QueryReport report) throws Exception {
+  public void send(@RequestBody() QueryReport report) throws Exception {
```
seanmcilvenna, line 382: `send` and `download` don't have any UI buttons right now. Further-more, they both expect `QueryReport`, which is no longer produced/used by the UI. In future tickets, we will want to change this signature to be `send(@RequestParam("reportId") String reportId)` or something like that.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -417,7 +392,7 @@ public void send (@RequestBody() QueryReport report) throws Exception {
   }
 
   @PostMapping("/download")
```
seanmcilvenna, line 394: We don't even have any requirements or UI mockups that make use of this. We may want to remove it. Didn't want to decide this, yet, though.

web/src/main/angular/app/app.component.ts
```diff
@@ -26,6 +26,5 @@ export class AppComponent implements OnInit {
   }
 
   async ngOnInit() {
-    await this.authService.loginLocal();
```
seanmcilvenna, line 29: Moved this to `app.module.ts` within the application's initialization as a whole. I think it's a more appropriate place for this type of logic.

web/src/main/angular/app/app.module.ts
```diff
@@ -11,14 +10,10 @@ import {OAuthModule} from 'angular-oauth2-oidc';
 import {AddHeaderInterceptor} from './auth-header-interceptor';
 import {ToastService} from './toast.service';
 import {ToastsContainerComponent} from './toasts-container/toasts-container.component';
-import {SmartLoginComponent} from './smart-login/smart-login.component';
```
seanmcilvenna, line 14: Removed stuff related to Smart-on-FHIR launching. Can bring it back later from history if we get a requirement for it.

web/src/main/angular/app/generate/generate.component.html
```diff
@@ -1,3 +1,36 @@
-<div *ngIf="authService.user">
-    <app-report-body></app-report-body>
+<div class="card">
```
seanmcilvenna, line 1: Moved stuff from `ReportBodyComponent` to here and removed the no-longer-valuable `ReportBodyComponent`

web/src/main/angular/app/generate/generate.component.ts
```diff
@@ -1,16 +1,100 @@
-import { Component, OnInit } from '@angular/core';
```
seanmcilvenna, line 1: Moved stuff from ReportBodyComponent to here and removed the no-longer-valuable ReportBodyComponent

web/src/main/angular/app/services/auth.service.ts
```diff
@@ -106,31 +56,35 @@ export class AuthService {
         if (loggedIn) {
           this.user = await this.oauthService.loadUserProfile() as any;
         }
-      } catch (ex) { }
+      } catch (ex) {
+        console.log(`Error loading user profile: ${ex.message}`);
```
seanmcilvenna, line 60: Useful to know why in the JS console

web/src/main/angular/app/services/auth.service.ts
```diff
@@ -106,31 +56,35 @@ export class AuthService {
...
       if (!this.user) {
         this.oauthService.initImplicitFlow(encodeURIComponent(this.router.url));
       } else {
+        console.log('Your token is: ' + this.token);
+        await this.oauthService.setupAutomaticSilentRefresh();
+
         this.token = this.oauthService.getIdToken();
+
         let path;
-        if(!this.oauthService.state || this.oauthService.state !== 'undefined'){
```
seanmcilvenna, line 117: Lots of little coding pattern differences...

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
arianamihailescu: Makes sense!

arianamihailescu (approved):

PR #90, opened 2021-08-16
NHSNLINK-520

seanmcilvenna:
Adding modal component that shows patient data for a given report
Adding button to view line-level data
Opening modal window when button is pressed
Stubbing out View/Exclude buttons for each patient (they don't do anything, yet)

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -480,37 +480,35 @@ public class ReportController extends BaseController {
       }
     }
 
-    if(!patientRequest.hasEntry() || patientRequest.getEntry().size() < 1){
-      throw new HttpResponseException(404, String.format("No patients in report with ID %s", id));
-    }
+    if (patientRequest.hasEntry()) {
+      Bundle patientBundle = client.transaction().withBundle(patientRequest).execute();
+      for (Bundle.BundleEntryComponent entry : patientBundle.getEntry()) {
+        PatientReportModel report = new PatientReportModel();
+        Patient patient = (Patient) entry.getResource();
 
-    Bundle patientBundle = client.transaction().withBundle(patientRequest).execute();
-    for(Bundle.BundleEntryComponent entry : patientBundle.getEntry()){
-      PatientReportModel report = new PatientReportModel();
-      Patient patient = (Patient) entry.getResource();
-
-      if(patient.getName().size() > 0){
-        if(patient.getName().get(0).getFamily() != null) {
-          report.setLastName(patient.getName().get(0).getFamily());
+        if (patient.getName().size() > 0) {
+          if (patient.getName().get(0).getFamily() != null) {
+            report.setLastName(patient.getName().get(0).getFamily());
+          }
+          if (patient.getName().get(0).getGiven().size() > 0 && patient.getName().get(0).getGiven().get(0) != null) {
+            report.setFirstName(patient.getName().get(0).getGiven().get(0).toString());
+          }
         }
-        if(patient.getName().get(0).getGiven().size() > 0 && patient.getName().get(0).getGiven().get(0) != null){
-          report.setFirstName(patient.getName().get(0).getGiven().get(0).toString());
+
+        if (patient.getBirthDate() != null) {
+          report.setDateOfBirth(Helper.getFhirDate(patient.getBirthDate()));
         }
-      }
 
-      if(patient.getBirthDate() != null){
-        report.setDateOfBirth(patient.getBirthDate().toString());
-      }
+        if (patient.getGender() != null) {
+          report.setSex(patient.getGender().toString());
+        }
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -480,37 +480,35 @@ public ReportModel getReport(
...
-
-    Bundle patientBundle = client.transaction().withBundle(patientRequest).execute();
-    for(Bundle.BundleEntryComponent entry : patientBundle.getEntry()){
-      PatientReportModel report = new PatientReportModel();
-      Patient patient = (Patient) entry.getResource();
-
-      if(patient.getName().size() > 0){
-        if(patient.getName().get(0).getFamily() != null) {
-          report.setLastName(patient.getName().get(0).getFamily());
+    if (patientRequest.hasEntry()) {
```
seanmcilvenna, line 483: UI would just crash if is thrown when no patients included in the report. Needed to modify API to just return an empty array when no patients were found, so that the UI didn't crash (or show an error to the user).

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -480,37 +480,35 @@ public ReportModel getReport(
...
-        report.setDateOfBirth(patient.getBirthDate().toString());
-      }
+        if (patient.getGender() != null) {
+          report.setSex(patient.getGender().toString());
+        }
 
-      if(patient.getGender() != null){
-        report.setSex(patient.getGender().toString());
-      }
+        if (patient.getId() != null) {
```
seanmcilvenna, line 506: `.getId()` returns "Patient/XXXX/_history/1" and I only wanted to respond with "XXXX".
`.getIdElement().getIdPart()` returns just the id part (i.e. "XXXX")

core/src/main/java/com/lantanagroup/link/Helper.java
```diff
@@ -26,11 +26,11 @@ public static String getFhirDate(LocalDateTime localDateTime) {
   }
 
   public static String getFhirDate(Date date) {
-    return new SimpleDateFormat("yyyy-MM-dd").format(date);
+    return new SimpleDateFormat(SIMPLE_DATE_FORMAT).format(date);
   }
   
   public static Date parseFhirDate(String dateStr) throws ParseException {
-	  return new SimpleDateFormat("yyyy-MM-dd").parse(dateStr);
+	  return new SimpleDateFormat(SIMPLE_DATE_FORMAT).parse(dateStr);
```
seanmcilvenna, line 33: Using fully specified dates, per previous discussions

web/src/main/angular/app/model/report-patient.ts
```diff
@@ -0,0 +1,12 @@
+export class ReportPatient {
+  id: string;
+  firstName: string;
+  lastName: string;
+  sex: string;
+  dateOfBirth: string;
+
+  getDOBDisplay() {
+    const date = new Date(Date.parse(this.dateOfBirth));
```
seanmcilvenna, line 9: Note: This can only be used in run-time when there is an _instance_ of the class. See the `ReportService.getReportPatients(reportId)` method for more info...

web/src/main/angular/app/services/report.service.ts
```diff
@@ -60,4 +62,17 @@ export class ReportService {
 
     saveAs(downloadResponse.body, fileName);
   }
+
+  async getReportPatients(reportId: string) {
+    const url = this.configService.getApiUrl(`/api/report/${encodeURIComponent(reportId)}/patient`);
+    return this.http.get<ReportPatient[]>(url)
+        // map the array of objects from the response to new instances of ReportPatient
+        .pipe(map(response => {
```
seanmcilvenna, line 70: You can't just do `return this.http.get<ReportPatient[]>(url)` because that just casts the objects in the response as ResponsePatient[] for the purpose of typescript/intellisense. You have to actually create instances of the JavaScript objects/classes and assign the properties of the HTTP response to the object. The objects in response from the HTTP request don't contain extra methods like `ReportPatient.getDOBDisplay()`.

web/src/main/angular/app/view-line-level/view-line-level.component.html
```diff
@@ -0,0 +1,43 @@
+<div class="modal-header">
+    <h4 class="modal-title" id="modal-basic-title">View Line-Level Data</h4>
+    <button type="button" class="close" aria-label="Close" (click)="activeModal.dismiss()">
+        <span aria-hidden="true">&times;</span>
+    </button>
+</div>
+<div class="modal-body">
+    <div class="alert alert-info" *ngIf="loading">Loading...</div>
```
seanmcilvenna, line 8: We should figure out how to standardize this "Loading..." concept. I think this is done differently on almost every single screen.

arianamihailescu (approved):

PR #91, opened 2021-08-16
NHSNLINK-445: Editing a calculated MeasureReport field

arianamihailescu:


web/src/main/angular/app/calculated-field/calculated-field.component.ts
```diff
@@ -68,6 +69,7 @@ export class CalculatedFieldComponent implements OnInit {
     } else if (value && reasonExt) {
       reasonExt.valueString = value;
     }
+    console.log(this.report);
```
seanmcilvenna, line 72: remove these console.log's please.

seanmcilvenna (changes requested):

PR #99, opened 2021-08-23
NHSNLINK-451: Fields are disabled/readonly when status is final

arianamihailescu:


web/src/main/angular/app/components/notes/notes.component.ts
```diff
@@ -34,26 +35,33 @@ export class NotesComponent implements OnInit {
...
             extensions.push(notesExt);
         } else {
             notesExtension.valueMarkdown = value;
         }
     }
 
-    async ngOnInit() {
+    get isDisabled() {
+        if (!this.report) return;
+        console.log("Status is:" + this.report.status);
```
seanmcilvenna, line 48: Please remove console.log

seanmcilvenna (changes requested):

PR #113, opened 2021-09-02
NHSNLINK-386 - Validation of access token used in REST requests

arianamihailescu:


web/src/main/angular/app/services/auth.service.ts
```diff
@@ -64,11 +64,8 @@ export class AuthService {
       if (!this.user) {
         this.oauthService.initImplicitFlow(encodeURIComponent(this.router.url));
       } else {
-        console.log('Your token is: ' + this.token);
-        await this.oauthService.setupAutomaticSilentRefresh();
-
         this.token = this.oauthService.getIdToken();
-
+        console.log('Your token is: ' + this.token);
```
seanmcilvenna, line 68: Remove this console.log please

web/src/main/angular/app/services/auth.service.ts
```diff
@@ -79,10 +76,17 @@ export class AuthService {
         if (path && path !== '/') {
           this.router.navigate([path]);
         }
+        await this.oauthService.setupAutomaticSilentRefresh();
```
seanmcilvenna, line 79: Did this get removed accidentally somewhere? Thought we had this already...

seanmcilvenna (changes requested):

web/src/main/angular/app/services/auth.service.ts
arianamihailescu: Removed

web/src/main/angular/app/services/auth.service.ts
arianamihailescu: Yes - but needed some changes around it to be sure we have the latest token.

PR #121, opened 2021-09-14
NHSNLINK-527: Detect changes before saving.

arianamihailescu:


web/src/main/angular/app/report/report.component.ts
```diff
@@ -39,18 +40,24 @@ export class ReportComponent implements OnInit, OnDestroy {
...
+
   viewLineLevel() {
     const modalRef = this.modal.open(ViewLineLevelComponent, {size: 'xl'});
     modalRef.componentInstance.reportId = this.reportId;
   }
 
   async send() {
     try {
-      if (confirm('Are you sure you want to submit? Changes will be saved before submitting...')) {
+      if (confirm('Before submitting, do you want to save your changes?')) {
```
seanmcilvenna, line 54: I think this should only show if the user has made changes... i.e.
```javascript
if (this.dirty) {
  if (confirm('Before submitting...')) {
    await this.save();
  } else {
    return;
  }
}

if (confirm('Are you sure ... submit?') {
...
```

seanmcilvenna (changes requested):

seanmcilvenna (approved):

PR #122, opened 2021-09-14
NHSNLINK-394

seanmcilvenna:
Allow patient identifier lists that use logical ID instead of identifier

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/ResourceIdChanger.java
+++ b/core/src/main/java/com/lantanagroup/link/ResourceIdChanger.java
@@ -75,6 +75,7 @@ public class ResourceIdChanger {
 
     // Find references that are invalid
     references.stream().filter(r -> {
+      if (r.getReference() == null) return false;
       String[] refParts = r.getReference().split("/");
       if (refParts.length != 2) return false;                 // Skip canonical references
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/lantanagroup/link/ResourceIdChanger.java
```diff
@@ -75,6 +75,7 @@ public void changeIds() {
 
     // Find references that are invalid
     references.stream().filter(r -> {
+      if (r.getReference() == null) return false;
```
seanmcilvenna, line 78: Found this bug while testing against http://test.sanerproject.org/fhir... Some elements of type `Reference` did not actually have a String reference specified in them. Spec definitely allows for that possibility, so considered this a bug fix.

core/src/main/java/com/lantanagroup/link/model/PatientOfInterestModel.java
```diff
@@ -0,0 +1,26 @@
+package com.lantanagroup.link.model;
+
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter @Setter
+public class PatientOfInterestModel {
```
seanmcilvenna, line 7: Creating a separate class to store information about a patient of interest, envisioning that this may expand in the future.

arianamihailescu (approved):

PR #128, opened 2021-09-22
NHSNLINK-551: Selecting Sep 8 returns for Sep 6

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -32,10 +37,12 @@ private MeasureReport generateMeasureReport() {
...
       logger.info(String.format("Executing $evaluate-measure for %s", this.context.getMeasureId()));
 
-      Parameters parameters = new Parameters();
-      parameters.addParameter().setName("periodStart").setValue(new DateType(this.criteria.getPeriodStart()));
-      parameters.addParameter().setName("periodEnd").setValue(new DateType(this.criteria.getPeriodEnd()));
+      Date startDate = Helper.parseFhirDate(this.criteria.getPeriodStart());
+      Date endDate = Helper.parseFhirDate(this.criteria.getPeriodEnd());
 
+      Parameters parameters = new Parameters();
+      parameters.addParameter().setName("periodStart").setValue(new InstantType(startDate, TemporalPrecisionEnum.SECOND, TimeZone.getDefault()));
```
seanmcilvenna, line 44: Your use of `InstantType` made me interested in whether or not the FHIR $evaluate-measure operation is supposed to _support_ an `instant`... Looking at the spec [here](https://www.hl7.org/fhir/operation-measure-evaluate-measure.html), I found that it lists the periodStart and periodEnd parameters as `date` types. Are you finding that HAPI is respecting the use of `InstantType`, or is it treating it as a `date`? That might be a source of our problems... We might _have_ to choose 12AM on the next day.

api/src/main/resources/application.yml
```diff
@@ -20,6 +20,8 @@ api:
     ventilator-codes-value-set: classpath:mechanical-ventilators.xml
     intubation-procedure-codes-value-set: classpath:intubation-procedures.xml
   document-reference-system: urn:ietf:rfc:3986
+  user:
```
seanmcilvenna, line 23: In future, we can probably refactor this to be `facility:` instead of `user:`
I think I found `user.timezone` referenced online, so that's why I mentioned it. I didn't necessarily mean we should create a `user` config section for this if it didn't exist... I thought it was already supported by SpringBoot.
Anyways, no need to fix in this PR (unless you want), but something to remember to refactor in the future.

seanmcilvenna (approved):

PR #131, opened 2021-10-01
NHSNLINK-557 - Search reports using periodStart and periodEnd

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/ApiInit.java
```diff
@@ -157,7 +161,32 @@ private void loadReportDefinitions() {
...
+      for (String filename : filenames) {
+        SearchParameter resource = readFileAsResource(xmlParser, filename);
+        fhirClient.update().resource(resource).execute();
+      }
+    } catch (Exception ex) {
+      logger.error(String.format("Error in loadSearchParameters due to %s", ex.getMessage()));
+    }
+  }
+
+  private SearchParameter readFileAsResource(IParser xmlParser, String fileName) {
```
seanmcilvenna, line 182: Should return `Resource` instead of `SearchParameter`. We may want to load other resource types in the future.

api/src/main/java/com/lantanagroup/link/api/ApiInit.java
```diff
@@ -157,7 +161,32 @@ private void loadReportDefinitions() {
     });
   }
 
+  private void loadSearchParameters() {
+    try {
+      String[] filenames = {"/fhir/document-reference-period-start-sp.xml",
```
seanmcilvenna, line 166: Can this list be: any file in the `/fhir` directory, instead of a code-defined list of files?

seanmcilvenna (changes requested):

PR #167, opened 2021-12-22
Nhsnlink 577

SaraMunich-Lantana:
I changed the name of the Query-API module to Agent and modified the pom and docker files. Please let me know if I need to change any other files I may have missed.


seanmcilvenna (changes requested):
Classes that start with `QueryXXX` should be renamed to `AgentXXX` (ex: `QueryApiApplication` renamed to `AgentApplication` and `QuerySecurityConfig` renamed to `AgentSecurityConfig` as well as others).
The `pom.xml` file contains a `build/finalName` property that should be updated to `agent`, and the `Dockerfile.agent` has the following line:
`COPY --from=build /tmp/query-api/target/link-query-api.war /usr/local/tomcat/webapps/ROOT.war`
This line needs to be updated to show `agent.war` to align with the `build/finalName`.

PR #173, opened 2022-01-03
Set up CI with Azure Pipelines

arianamihailescu:


seanmcilvenna:
Would need @michael-misiaszek to review this to determine if it can be supported by our azure devops system. At a glance, looks like the devops pipeline would be stored in the GitHub repo instead of in our DevOps environment. Not sure there's much value-add here for _us_, but probably is for external people that develop/build Link.

PR #175, opened 2022-01-06
NHSNLINK-617: Refactor ApiInit.java to use FHIRDataProvider

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/controller/UserController.java
```diff
@@ -28,7 +28,7 @@
   @GetMapping
   public List<UserModel> getUsers (Authentication authentication, HttpServletRequest request) throws Exception {
 
-    IGenericClient fhirStoreClient = this.getFhirStoreClient();
+    IGenericClient fhirStoreClient = this.getFhirStoreProvider().getClient();
```
seanmcilvenna, line 31: Nice temporary step until we refactor UserController

seanmcilvenna (approved):

PR #181, opened 2022-01-11
Replaced the getFhirStoreClient with getFhirDataProvider() in UserCon…

SaraMunich-Lantana:
I created the searchPractitioner(String tagSystem, String tagValue) in FhirDataProvider and used it in UserController to replace fhirStoreClient.search()...

I also created a UnitTest for this, which had passed.


api/src/test/java/com/lantanagroup/link/api/controller/UserControllerTest.java
```diff
@@ -0,0 +1,23 @@
...
+  private String tagValueTest = "link-user";
+
+  @Test
+  public void getUsers() {
+
+    FhirDataProvider fhirDataProvider = mock(FhirDataProvider.class);
+    Bundle bundle = fhirDataProvider
+            .searchPractitioner(tagSystemTest, tagValueTest);
+  }
+}
```
arianamihailescu, line 23: Can we assert something here? Like we get a count of resources > 0 for example. 
Also can we add a negative test in which the tag system or value have invalid values and then assert that the count of resources in the bundle is 0.

seanmcilvenna (approved):

PR #190, opened 2022-01-27
Nhsnlink 591

arianamihailescu:


core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -26,6 +27,23 @@
   private static final String SUBJECT = "sub";
   private static final String DOCUMENT_REFERENCE_VERSION_URL = "https://www.cdc.gov/nhsn/fhir/nhsnlink/StructureDefinition/nhsnlink-report-version";
 
+  public static void recordAuditEvent(HttpServletRequest request, IGenericClient fhirClient, DecodedJWT jwt, AuditEventTypes type, String outcomeDescription) {
```
seanmcilvenna, line 30: The use of IGenericClient shouldn't be added here, and should be put in FhirDataProvider

core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -357,37 +371,4 @@ public static Bundle getBundle(String content) {
...
   }
-
-  /**
-   * Creates a Bundle of type "batch" with an entry for each resource provided. Expects that each resource
-   * have an id, so that it can create a "request" with a method of "PUT resourceType/id"
-   * Example: createUpdateBatch(List.of(resource1, resource2, resource3))
-   * @param resources The list of resources to be added to the batch
-   * @return Bundle that can be executed as a transaction on the FHIR server
-   */
-  public static Bundle createUpdateBatch(List<DomainResource> resources) {
```
seanmcilvenna, line 368: This is a potentially useful method, but I don't think is actually being used anywhere, yet.
Suggest we restore this method and consider it for use in the future, particularly around creating transaction bundles to update multiple resources in one request.

seanmcilvenna (changes requested):

PR #241, opened 2022-03-11
Nhsnlink 759

SaraMunich-Lantana:
Created the QueryResponse class to contain Patient Id and a Patient data bundle, then modified Query.execute, ReportController.getRemotePatientData, ReportController.queryAndStorePatientData, and AgentController.getData to return List<QueryResponse>.  I then modified ReportContext to store a List<QueryResponse> as patientData and modified GenerateResponse to store the List<QueryResponse> generated into ReportContext.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -210,7 +210,7 @@ public class ReportController extends BaseController {
    * @return Returns a list of the logical ids for the Patient resources stored on the internal fhir server
    * @throws Exception
    */
-  private List<String> queryAndStorePatientData(List<PatientOfInterestModel> patientsOfInterest) throws Exception {
+  private List<QueryResponse> queryAndStorePatientData(List<PatientOfInterestModel> patientsOfInterest) throws Exception {
     try {
       Bundle patientDataBundle = null;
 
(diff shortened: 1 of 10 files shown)


query/src/main/java/com/lantanagroup/link/query/uscore/Query.java
```diff
@@ -10,42 +11,46 @@
...
+      try {
+        PatientScoop scoop = this.applicationContext.getBean(PatientScoop.class);
+        scoop.setFhirQueryServer(this.getFhirQueryClient());
+        scoop.execute(patientsOfInterest);
+
+        List<PatientData> patientDatas = scoop.getPatientData();
+
+        for (PatientData patientData : patientDatas) {
+          Bundle patientBundle = patientData.getBundleTransaction();
+          QueryResponse queryResponse = new QueryResponse(patientBundle.getEntry().get(0).getResource().getIdElement().getIdPart(), patientBundle);
```
seanmcilvenna, line 42: Is it gauranteed that `patientBundle.getEntry().get(0).getResource()` is always the `Patient` resource?
If not, consider doing a `patientBundle.getEntry().filter(... resource of type Patient ...).findOne()` call first and logging an error (and skipping) if a Patient resource wasn't found.

core/src/main/java/com/lantanagroup/link/model/QueryResponse.java
```diff
@@ -0,0 +1,21 @@
+package com.lantanagroup.link.model;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.hl7.fhir.r4.model.Bundle;
+
+@Getter
+@Setter
+public class QueryResponse {
+  String patientId;
```
seanmcilvenna, line 10: Please explicitly declare variables like this as `private`

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -210,66 +225,65 @@ private Bundle getRemotePatientData(List<PatientOfInterestModel> patientsOfInter
...
+        // Fix resource IDs in the patient data bundle that are invalid (longer than 64 characters)
+        // (note: this also fixes the references to resources within invalid ids)
+        ResourceIdChanger.changeIds(patientQueryResponse.getBundle());
+
+        // For debugging purposes:
+        //String patientDataBundleXml = this.ctx.newXmlParser().encodeResourceToString(patientDataBundle);
+
+        // Store the data
+        logger.info("Storing data for the patients: " + StringUtils.join(patientsOfInterest, ", "));
+        Bundle response = this.getFhirDataProvider().transaction(patientQueryResponse.getBundle());
```
seanmcilvenna, line 264: In the future, this line will change to _store_ the Bundle, rather than execute the bundle as a transaction.
For now, this is needed because $evaluate-measure on HAPI looks at the resources _on_ the server itself to evaluate. But, when we move to use CQF-Ruler's $evaluate-measure, we will be passing the Bundle of patient data to the CQF-Ruler for each patient. So, that's why we will change this to _store_ the bundle, rather than execute the bundle as a batch/transaction. Nothing to do here, just an FYI for anticipating future changes.

seanmcilvenna (approved):
Please review comments. Don't have any specific changes to request. Just a couple things to consider. Feel free to merge and move this align to "QA". I spent about 20 minutes reviewing the changes. Nice work!

PR #250, opened 2022-03-21
CQF Ruler Integration

SaraMunich-Lantana:
Added the patientDataBundle from QueryResponse to Paramters for getting the master measure report as "additionalData".

Merging into new branch for epic 672 https://lantana.atlassian.net/browse/LINK-672


seanmcilvenna (changes requested):
I suggest renaming this branch to "cqf-ruler-integration" (if that's even possible), and using it as the basis for the entire Epic.
I think committing these changes back to master will _potentially_ break things.
For example, if we merge these changes into master without having LINK-760 done, then it will attempt to evaluate the measure on the evaluation service, but the evaluation service won't have the measure resources on it _to_ evaluate, and will just return an error.

PR #251, opened 2022-03-22
Nhsnlink 760

SaraMunich-Lantana:
Created a new instance of FhirDataProvider that points to the measure evaluation service base URL to store the measure resources in storeReportBundleResources and modified the FhirDataProvider's constructor to take in a string.

Doing this before 660 hoping it helps prevent things from breaking.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -95,11 +95,11 @@ public class ReportController extends BaseController {
       }
     });
 
-    logger.info("Executing the measure definition bundle as a transaction on " + this.config.getFhirServerStore());
+    logger.info("Executing the measure definition bundle");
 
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -96,7 +96,12 @@ private String storeReportBundleResources(Bundle bundle) {
 
     logger.info("Executing the measure definition bundle as a transaction on " + this.config.getFhirServerStore());
 
-    this.getFhirDataProvider().transaction(bundle);
+    if(this.config.getEvaluationService() == null) {
```
seanmcilvenna, line 99: Shouldn't hard-code this...
Throw an exception instead. A ConfigurationException (or something like that?)
Configuration _must_ have 'evaluation-service' defined in it.

seanmcilvenna (changes requested):

PR #264, opened 2022-04-11
CLI adjustments to calculate reporting period for generate-and-submit

SaraMunich-Lantana:
Updated GenerateAndSubmitConfig and GenerateAndSubmitCommand to support period-start: adjust-day, adjust-month, and start-of-day as well as period-end: adjust-day and end-of-day.


seanmcilvenna (changes requested):
Generally speaking, looks good. Would you please create some unit tests for this functionality, though?

PR #266, opened 2022-04-13
Link 729

SaraMunich-Lantana:
Moved txServiceFilter to FhirHelper and used it in ApiInit.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -98,12 +98,35 @@ public class ReportController extends BaseController {
     logger.info("Executing the measure definition bundle");
 
     // Store the resources of the measure on the evaluation service
+    bundle = txServiceFilter(bundle);
     FhirDataProvider fhirDataProvider = new FhirDataProvider(this.config.getEvaluationService());
     fhirDataProvider.transaction(bundle);
 
     logger.info("Done executing the measure definition bundle");
 
     return measureId;
+  }
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -439,5 +440,29 @@ public static List<MeasureReport> getPatientReports(List<String> patientMeasureR
             .map(e -> (MeasureReport) e.getResource())
             .collect(Collectors.toList());
   }
+
+  public static Bundle txServiceFilter(Bundle bundle, ApiConfig config) {
```
seanmcilvenna, line 444: Suggest renaming to `storeTerminologyAndReturnOther(Bundle, ApiConfig)`
Also, please provide a javadoc description to this method about what it does.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -95,37 +95,14 @@ private void storeReportBundleResources(Bundle bundle, ReportContext context) {
     logger.info("Executing the measure definition bundle");
 
     // Store the resources of the measure on the evaluation service
-    bundle = txServiceFilter(bundle);
     bundle.setType(Bundle.BundleType.TRANSACTION);
```
seanmcilvenna, line 98: Suggest changing this to BATCH instead of TRANSACTION.
We're using BATCH for terminology, which will likely run a lot faster.

seanmcilvenna (changes requested):

PR #269, opened 2022-04-15
Link-775: Extract list of resource types from Library.dataRequirements

arianamihailescu:


core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -474,5 +473,19 @@ public static Bundle storeTerminologyAndReturnOther(Bundle bundle, ApiConfig con
     }
     return bundle;
   }
+
+  public static List getDataRequirementTypes(Bundle reportRefBundle) {
+    Optional<Library> foundLibrary = reportRefBundle.getEntry().stream()
```
seanmcilvenna, line 478: Rather than `findFirst()`, suggest you look for all Library resources in the Bundle. There may be multiple Library resources, and one Library may not have the same dataRequirements as another Library. Will need to build a List<Library> and extract the data requirements for all of them, and then return a unique list of resource types.

seanmcilvenna (changes requested):

PR #272, opened 2022-04-20
LINK-670

amphillipsLGC:


core/src/main/java/com/lantanagroup/link/auth/LinkAuthManager.java
```diff
@@ -11,8 +11,11 @@
 
 public class LinkAuthManager implements AuthenticationManager {
```
seanmcilvenna, line 12: Consider using `@RequiredArgsConstructor` from lombok. Removes the need to explicitly declare the constructor for private @NonNull fields. Not a requirement... just something to consider.
https://projectlombok.org/features/constructor

seanmcilvenna (approved):
Feel free to merge and delete the PR after considering and/or addressing the comments in the review; I don't need to re-review. Overall, looks great!

PR #273, opened 2022-04-20
Link 767

SaraMunich-Lantana:
Added requireHttps as a boolean to QueryConfig, created AgentInit and AgentConfig, and modified AgentInit, ApiInit, AgentController, and QueryCommand to validate that query URLs meet the “HTTPS” requirement.


agent/src/main/java/com/lantanagroup/link/agent/AgentInit.java
```diff
@@ -0,0 +1,19 @@
...
+
+  @Autowired
+  private AgentConfig config;
+
+  @Autowired
+  private QueryConfig queryConfig;
+
+  private void init() {
+    if (this.queryConfig.isRequireHttps()) {
+      // TODO: Validate HTTPS for URLs in queryConfig
```
seanmcilvenna, line 16: Think this should be populated with the change you have in `AgentController` right now... Would prefer the agent to fail to startup if the config requires HTTPS but is not actually configured with HTTPS urls, rather than waiting for a query request in the controller to execute that validation

api/src/main/java/com/lantanagroup/link/api/ApiInit.java
```diff
@@ -59,6 +58,11 @@ private void loadMeasureDefinitions() {
               .uri(URI.create(measureDefUrl))
               .build();
 
+      if (queryConfig.isRequireHttps() && !measureDefUrl.contains("https")) {
```
seanmcilvenna, line 61: Maybe should be `measureDefUrl.startsWith("https://")` instead?

cli/src/main/java/com/lantanagroup/link/cli/QueryCommand.java
```diff
@@ -38,6 +39,10 @@ public void query(String patient, @ShellOption(defaultValue = "") String output)
       this.registerBeans();
 
       QueryConfig config = this.applicationContext.getBean(QueryConfig.class);
+      if(config.isRequireHttps() && !config.getFhirServerBase().contains("https")) {
```
seanmcilvenna, line 42: `startsWith("https://")` ??

seanmcilvenna (changes requested):

seanmcilvenna:
Go ahead and merge your changes and close the ticket/pR out after reviewing/addressing my comments.

PR #278, opened 2022-04-28
LINK-570: Upgrade Angular to version 13 stable

smailliwcs:


web/src/main/angular/app/services/auth.service.ts
```diff
@@ -53,7 +53,8 @@ export class AuthService {
 
       try {
         if (loggedIn && this.oauthService.hasValidAccessToken()) {
-          this.user = await this.oauthService.loadUserProfile() as any;
+          let userProfile: any = await this.oauthService.loadUserProfile();
```
smailliwcs, line 56: The structure of the object returned by `OAuthService.loadUserProfile` seems to have changed between v10 and v12 of `angular-oauth2-oidc`.  We now need to access the user's information through the `info` property.  I couldn't find any documentation on this, so I added a check below that falls back to the old behavior if the `info` property doesn't exist.

seanmcilvenna (approved):

PR #279, opened 2022-04-28
Send configured terminology service to CQF-Ruler during $evaluate-measure

SaraMunich-Lantana:
Checked to see if ApiConfig.terminologyService and ApiConfig.evaluationService are the same in MeasureEvaluator, if not then created a new endpoint set as active with the Address as the same as the ApiConfig.terminologyService and a connectionType with "http://terminology.hl7.org/CodeSystem/endpoint-connection-type" as its system and "hl7-fhir-rest" as its code. This endpoint was then added as a parameter for getting the measure report.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
+++ b/api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
@@ -48,6 +48,16 @@ public class MeasureEvaluator {
       parameters.addParameter().setName("periodEnd").setValue(new StringType(this.criteria.getPeriodEnd().substring(0, this.criteria.getPeriodEnd().indexOf("."))));
       parameters.addParameter().setName("subject").setValue(new StringType(patientId));
       parameters.addParameter().setName("additionalData").setResource(patientData.getBundle());
+      if(this.config.getEvaluationService() != this.config.getTerminologyService()) {
+        Endpoint terminologyEndpoint = new Endpoint();
+        terminologyEndpoint.setStatus(Endpoint.EndpointStatus.ACTIVE);
+        terminologyEndpoint.setConnectionType(new Coding());
+        terminologyEndpoint.getConnectionType().setSystem("http://terminology.hl7.org/CodeSystem/endpoint-connection-type");
+        terminologyEndpoint.getConnectionType().setCode("hl7-fhir-rest");
+        terminologyEndpoint.setAddress(this.config.getTerminologyService());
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -48,6 +48,16 @@ private MeasureReport generateMeasureReport() {
       parameters.addParameter().setName("periodEnd").setValue(new StringType(this.criteria.getPeriodEnd().substring(0, this.criteria.getPeriodEnd().indexOf("."))));
       parameters.addParameter().setName("subject").setValue(new StringType(patientId));
       parameters.addParameter().setName("additionalData").setResource(patientData.getBundle());
+      if(this.config.getEvaluationService() != this.config.getTerminologyService()) {
```
seanmcilvenna, line 51: In Java, I don't believe you can do `string != string`
I believe it has to be `!string.equals(string)`
Suggest setting up a unit test to confirm this, which would be beneficial for the functionality anyways.

api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -48,6 +48,16 @@ private MeasureReport generateMeasureReport() {
...
+        Endpoint terminologyEndpoint = new Endpoint();
+        terminologyEndpoint.setStatus(Endpoint.EndpointStatus.ACTIVE);
+        terminologyEndpoint.setConnectionType(new Coding());
+        terminologyEndpoint.getConnectionType().setSystem("http://terminology.hl7.org/CodeSystem/endpoint-connection-type");
```
seanmcilvenna, line 55: Move this hard-coded string to `Constants` class please

api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -48,6 +48,16 @@ private MeasureReport generateMeasureReport() {
...
+        terminologyEndpoint.getConnectionType().setCode("hl7-fhir-rest");
```
seanmcilvenna, line 56: ...and this, maybe just as a static class field in MeasureEvaluator... I see this being less valuable in the `Constants` class.

seanmcilvenna (changes requested):

seanmcilvenna (approved):

PR #293, opened 2022-05-17
Remove report resources after submitting

SaraMunich-Lantana:
Added functionality to delete resources then deleted patient data bundles, patient measure reports, the patient census list, and the master measure report after submitting a generated report.


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -471,11 +452,51 @@ public void send(
...
+
+      List<ListResource> census = FhirHelper.getCensusLists(documentReference, this.getFhirDataProvider());
+      for(ListResource.ListEntryComponent entry: census.get(0).getEntry()) {
+        if(entry.getItem().getReference() != null) {
+          String patientRef = entry.getItem().getReference().contains("Patient/")?
+                  entry.getItem().getReference().substring("Patient/".length()):entry.getItem().getReference();
+          String patientReportID = String.valueOf(patientRef.hashCode());
+
+          try {
+            this.getFhirDataProvider().deleteResource("Bundle", patientReportID, true);
```
arianamihailescu, line 481: I suppose the "Bundle" is the patient bundle. In that case the id should be:  masterMeasureReportID + "-" + patientReportID?

arianamihailescu: Consider that could be multiple census lists for 1 report so the code should delete all of them and loop through all of them to delete the associated patient bundles, reports etc.

core/src/main/java/com/lantanagroup/link/FhirDataProvider.java
```diff
@@ -321,4 +324,15 @@ public Bundle searchBundleByTag(String system, String value) {
             .execute();
   }
 
+  public void deleteResource(String resourceType, String id, boolean permanent) {
+    try {
+      URL url = new URL(this.client.getServerBase() + "/" + resourceType + "/" + id + (permanent?"?_expunge=true":""));
+      HttpURLConnection con = (HttpURLConnection) url.openConnection();
+      con.setRequestMethod("DELETE");
+      con.getResponseMessage();
+      con.disconnect();
```
arianamihailescu, line 333: Can we use IGenericClient delete operation instead of HttpUrlConnection and also delete all resources of the same type at once (like in a transaction operation)? Might be more performant.

SaraMunich-Lantana: The reason it uses an http request to delete is because Sean said that IGenericClient doesn't support the option for "?_expunge=true" yet
 and that IGenericClient would be refactored to support it down the road.

arianamihailescu (approved):

PR #303, opened 2022-06-02
Retrieve census from submission server when re-generating

SaraMunich-Lantana:
Created functionality to retrieve the patient census list from the submission bundle that was retrieved from the sent location that was added to the doc reference during the last submission. The census list is then added to the dev server and the report context where it is used to populate a list of patients of interest instead of looking them up from the server.

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/IReportSender.java
+++ b/core/src/main/java/com/lantanagroup/link/IReportSender.java
@@ -1,5 +1,9 @@
 package com.lantanagroup.link;
 
+import ca.uhn.fhir.context.FhirContext;
+import com.lantanagroup.link.config.api.ApiConfig;
+import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.DocumentReference;
 import org.hl7.fhir.r4.model.MeasureReport;
 import org.springframework.security.core.Authentication;
 
@@ -7,4 +11,6 @@ import javax.servlet.http.HttpServletRequest;
 
 public interface IReportSender {
   void send(MeasureReport masterMeasureReport, HttpServletRequest request, Authentication auth, FhirDataProvider fhirDataProvider, Boolean sendWholeBundle, boolean removeGeneratedObservations) throws Exception;
+
+  Bundle retrieve(ApiConfig config, FhirContext fhirContext, DocumentReference existingDocumentReference);
 }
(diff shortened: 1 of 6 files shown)


nhsn/src/main/java/com/lantanagroup/link/nhsn/JSONSender.java
```diff
@@ -26,6 +29,11 @@ public void send(MeasureReport masterMeasureReport, HttpServletRequest request,
     FhirHelper.recordAuditEvent(request, fhirDataProvider, ((LinkCredentials) auth.getPrincipal()).getJwt(), FhirHelper.AuditEventTypes.Send, "Successfully sent report");
   }
 
+  @Override
+  public Bundle retrieve(ApiConfig config, FhirContext fhirContext, DocumentReference existingDocumentReference) {
+    return null;
```
seanmcilvenna, line 34: If we swap out FHIRSender in config to JSONSender, returning null here will break the system.
1. If null is returned, does it ReportController handle that? i.e. log a message saying "Could not retrieve census to re-generate report" or something like that?
2. Can we setup a base-class for the senders in `core` that allows us to re-use the same `retrieve()` logic between FHIRSender and JSONSender?

thsa/src/main/java/com/lantanagroup/link/thsa/CSVSender.java
```diff
@@ -28,6 +31,11 @@ public void send(MeasureReport masterMeasureReport, HttpServletRequest request,
     FhirHelper.recordAuditEvent(request, fhirDataProvider, ((LinkCredentials) auth.getPrincipal()).getJwt(), FhirHelper.AuditEventTypes.Send, "Successfully sent report");
   }
 
+  @Override
+  public Bundle retrieve(ApiConfig config, FhirContext fhirContext, DocumentReference existingDocumentReference) {
+    return null;
```
seanmcilvenna, line 36: Same question as above about base class?

core/src/main/java/com/lantanagroup/link/IReportSender.java
```diff
@@ -1,10 +1,16 @@
...
+import org.hl7.fhir.r4.model.DocumentReference;
 import org.hl7.fhir.r4.model.MeasureReport;
 import org.springframework.security.core.Authentication;
 
 import javax.servlet.http.HttpServletRequest;
 
 public interface IReportSender {
   void send(MeasureReport masterMeasureReport, HttpServletRequest request, Authentication auth, FhirDataProvider fhirDataProvider, Boolean sendWholeBundle, boolean removeGeneratedObservations) throws Exception;
+
+  Bundle retrieve(ApiConfig config, FhirContext fhirContext, DocumentReference existingDocumentReference);
```
seanmcilvenna, line 15: Please put a comment on this to help other developers understand how `retrieve()` is used in an instance of `IReportSender`

seanmcilvenna (changes requested):

seanmcilvenna (approved):

PR #305, opened 2022-06-02
Extend Coding used by ConceptMapping to preserve original value before translation

SaraMunich-Lantana:
I modified ApplyConceptMaps to add an extension with the original code to the code before the code gets its new values from the concept map.

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/nhsn/ApplyConceptMaps.java
+++ b/core/src/main/java/com/lantanagroup/link/nhsn/ApplyConceptMaps.java
@@ -7,6 +7,7 @@ import com.lantanagroup.link.model.ReportContext;
 import com.lantanagroup.link.model.ReportCriteria;
 import org.hl7.fhir.r4.model.Coding;
 import org.hl7.fhir.r4.model.ConceptMap;
+import org.hl7.fhir.r4.model.Extension;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/lantanagroup/link/nhsn/ApplyConceptMaps.java
```diff
@@ -18,11 +19,16 @@ public class ApplyConceptMaps implements IReportGenerationEvent {
...
 
-  private void getConvertCode(ConceptMap map, Coding code) {
+  private void applyMap(ConceptMap map, Coding code) {
     // TODO: Lookup ConceptMap.group based on code system
     map.getGroup().stream().forEach((ConceptMap.ConceptMapGroupComponent group) -> {
       if (group.getSource().equals(code.getSystem())) {
         List<ConceptMap.SourceElementComponent> elements = group.getElement().stream().filter(elem -> elem.getCode().equals(code.getCode())).collect(Collectors.toList());
+        // preserve original code
+        Extension originalCode = new Extension();
+        originalCode.setUrl("https://www.lantanagroup.com/fhir/StructureDefinition/mapped-concept");
```
seanmcilvenna, line 29: Please put this constant in `Constants` within the `core` module

core/src/main/java/com/lantanagroup/link/nhsn/ApplyConceptMaps.java
```diff
@@ -18,11 +19,16 @@ public class ApplyConceptMaps implements IReportGenerationEvent {
...
+        originalCode.setValue(code);
```
seanmcilvenna, line 30: Should this be `code.copy()`? 
When `setSystem()` and `setDisplay()` and `setCode()` get called after this, won't the instance used by the extension _also_ get updated?

seanmcilvenna (changes requested):
Since this is such targeted code without interaction with a FHIR server, please make sure to add unit tests for the logic.

seanmcilvenna (approved):

PR #311, opened 2022-06-10
Update ReportController methods to use individual patient bundles

SaraMunich-Lantana:
I created functionality for ReportController.getPatientBundles() to return a list of individual patient bundles (with the option to specify for a single patient or patient report) to be used in getReportPatients and getPatientData


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -1166,6 +1031,166 @@ public ReportModel excludePatients(
     return report;
   }
 
+  /**
+   * Retrieves patient data bundles either from a submission bundle if its report had been sent
```
arianamihailescu, line 1035: This method seems to be very long and have duplicated code - for readability can we make it so that we can use the same code but pass different lists to loop through?

arianamihailescu (approved):

PR #315, opened 2022-06-15
LINK-960

amphillipsLGC:
Add ability for PatientIdentifierController to search by rerefernce as well as identifier. Prioritizing searching by reference.


SaraMunich-Lantana (approved):
Looks good :)

PR #318, opened 2022-06-20
LINK-874: Transform CSV report submission in consumer and store

smailliwcs:
Note that the code added to `thsa` is from `NHSNLINK-694` (and ultimately from [Saner-CSV-Converter](https://github.com/AudaciousInquiry/Saner-CSV-Converter)).  It should not need to be reviewed closely.


consumer/src/main/java/com/lantanagroup/link/consumer/api/ReportCsvOperationProvider.java
```diff
@@ -0,0 +1,200 @@
...
+  /**
+   * Returns the measure identified by the specified ID.
+   * First, attempts to read the measure from this server.
+   * If that fails, attempts to parse a Java resource from <code>consumer</code> or <code>thsa</code>.
+   */
+  private Measure getMeasure(IdType measureId, RequestDetails requestDetails) {
+    try {
+      logger.debug("Requesting measure from server");
+      return measureDao.read(measureId, requestDetails);
+    } catch (ResourceNotFoundException | ResourceGoneException restException) {
```
arianamihailescu, line 79: Might be good to load the Measure when the application starts up.

arianamihailescu (changes requested):

arianamihailescu (approved):

PR #335, opened 2022-07-05
LINK-995: Create configuration for Epic Patient List update interval

smailliwcs:


smailliwcs:
Before pushing to `master`, the dev config for `api` should be updated so that the items under `api.report-defs.urls` are converted from bare strings to mappings with the key `url`:

```yml
api:
  report-defs:
    urls:
      - url: https://ehr-test.nhsnlink.org/fhir/Bundle/InitialInpatientPopulation
      - url: https://ehr-test.nhsnlink.org/fhir/Bundle/THSAMeasure
```

PR #336, opened 2022-07-05
Treat submission server (report consumer) as FHIR server

SaraMunich-Lantana:
I created functionality to submit to and retrieve from a server as a FHIR server in FhirDataProvider with gzip and authentication support as interceptors.


smailliwcs:
Maybe I'm just not seeing it here in GitHub, but has `api` been updated to use these new methods during submission and post-submission retrieval?

amphillipsLGC:
It appears it should be replacing logic where the following are used: IGenericClient.create(), IGenericClient.update() and IGenericClient.retrieve() 

Maybe a new ticket to update the API calls would make sense.

PR #341, opened 2022-07-11
Update configuration and generic create/update/retrieve methods to use new FhirDataProvider methods

SaraMunich-Lantana:
I integrated FhirDataProvider.submitToServer() into GenericSender and FhirDataProvider.retrieveFromServer into FHIRReceiver. Then I merged JSONSender, XMLSender, and CSVSender into a single FHIRSender class.


core/src/main/java/com/lantanagroup/link/FhirDataProvider.java
```diff
@@ -342,27 +346,36 @@ public Bundle searchBundleByTag(String system, String value) {
...
     IGenericClient submissionClient = this.ctx.newRestfulGenericClient(serverBase);
     submissionClient.registerInterceptor(new GZipContentInterceptor());
-    IClientInterceptor authInterceptor = new BasicAuthInterceptor(senderConfig.getUsername(), senderConfig.getPassword());
+    BearerTokenAuthInterceptor authInterceptor = null;
+    try {
+      authInterceptor = new BearerTokenAuthInterceptor(token);
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
     submissionClient.registerInterceptor(authInterceptor);
```
smailliwcs, line 358: This will throw a `NullPointerException` if the above constructor for `BearerTokenAuthInterceptor` failed.  Not sure why we're catching that exception.  Do we need to allow `token` to be null, and *not* add the interceptor in that case?  If so, I would recommend removing the `try`/`catch` and replacing it with an explicit null check for `token`.  Otherwise we should probably let the exception bubble up to the caller.

nhsn/src/main/java/com/lantanagroup/link/nhsn/FHIRSender.java
```diff
@@ -17,19 +14,34 @@
...
+      return fhirDataProvider.bundleToXml(bundle);
+    }
+
+    if(type.equals("csv")) {
+      // TODO: Use Keith's CSV conversion code to convert Bundle to CSV
+      String csv = "";
+      return csv;
+    }
+
+    return "";
```
smailliwcs, line 45: Recommend throwing an exception here since the caller passed in a `type` we don't support.

smailliwcs (changes requested):

PR #345, opened 2022-07-15
Measure storage operation

SaraMunich-Lantana:
I created functionality and a unit test for MeasureStoreController.storeMeasure() to store measure bundles in the current working directory within the path specified in MeasureEvalConfig.


measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreController.java
```diff
@@ -25,7 +39,24 @@ public void initBinder(WebDataBinder binder) {
 
   @PostMapping("/$store-measure")
   public void storeMeasure(@RequestBody Bundle measureBundle) {
-    // TODO: Store Bundle on file system in a file named <measureBundle.id>.xml
+    FhirContext ctx = FhirContextProvider.getFhirContext();
+    String measureContentXML = ctx.newXmlParser().encodeResourceToString(measureBundle);
```
seanmcilvenna, line 43: Should null/empty-check the `measureBundle.getId()`, since it is used to populate the filename, and therefore required.

measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreController.java
```diff
@@ -25,7 +39,24 @@ public void initBinder(WebDataBinder binder) {
...
+    try(BufferedWriter writer = Files.newBufferedWriter(
+            Paths.get(this.config.getMeasuresPath() + File.separator + measureBundle.getId() + ".xml"),
+            StandardCharsets.UTF_8, StandardOpenOption.CREATE)) {
+      writer.write(measureContentXML);
+      writer.flush();
+    }catch(IOException ex)
+    {
+      ex.printStackTrace();
+    }
     System.out.println(this.config.getMeasuresPath());
```
seanmcilvenna, line 60: Remove `System.out.println()` and replace with `logger.debug()` if appropriate.

measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreControllerTest.java
```diff
@@ -0,0 +1,27 @@
...
+
+import static org.junit.Assert.*;
+
+public class MeasureStoreControllerTest {
+
+  @Test
+  public void storeMeasure() {
+    Bundle measureBundle = new Bundle();
+    measureBundle.setId("UnitTest");
+    MeasureEvalConfig config = new MeasureEvalConfig();
```
seanmcilvenna, line 20: Don't necessarily need to make a change here... Just a thought/consideration:
Should unit tests write to file systems? I know best practices indicate that unit tests shouldn't write to DBs... I kinda consider file systems to be one in the same.
I wonder if there should be a mechanism to mock out the process of writing to a file so that we can assume writing to a file works in the unit test, without actually modifying the file system?
Perhaps consider creating a separate method that uses the simplified `Files` class (see https://www.baeldung.com/java-write-to-file section 8) to write/create the file and then mock that method out in the unit test to assume it succeeds?

seanmcilvenna (changes requested):

measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreController.java
```diff
@@ -25,7 +39,24 @@ public void initBinder(WebDataBinder binder) {
...
   @PostMapping("/$store-measure")
   public void storeMeasure(@RequestBody Bundle measureBundle) {
-    // TODO: Store Bundle on file system in a file named <measureBundle.id>.xml
+    FhirContext ctx = FhirContextProvider.getFhirContext();
+    String measureContentXML = ctx.newXmlParser().encodeResourceToString(measureBundle);
+    if (!Files.exists(Paths.get(this.config.getMeasuresPath()))) {
+      try {
+        Files.createDirectories(Paths.get(this.config.getMeasuresPath()));
+      } catch (IOException e) {
+        e.printStackTrace();
```
arianamihailescu, line 48: Let the IO Exception be thrown  - they will be handled in the Global exception handler.

measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreController.java
```diff
@@ -25,7 +39,24 @@ public void initBinder(WebDataBinder binder) {
...
+      }
+    }
+    try(BufferedWriter writer = Files.newBufferedWriter(
+            Paths.get(this.config.getMeasuresPath() + File.separator + measureBundle.getId() + ".xml"),
+            StandardCharsets.UTF_8, StandardOpenOption.CREATE)) {
+      writer.write(measureContentXML);
+      writer.flush();
+    }catch(IOException ex)
+    {
+      ex.printStackTrace();
```
arianamihailescu, line 58: Let the IO Exception be thrown - they will be handled in the Global exception handler.

measureeval/src/main/java/com/lantanagroup/link/measureeval/controller/MeasureStoreController.java
```diff
@@ -24,8 +42,17 @@ public void initBinder(WebDataBinder binder) {
...
   @PostMapping("/$store-measure")
-  public void storeMeasure(@RequestBody Bundle measureBundle) {
-    // TODO: Store Bundle on file system in a file named <measureBundle.id>.xml
-    System.out.println(this.config.getMeasuresPath());
+  public void storeMeasure(@RequestBody Bundle measureBundle) throws IOException {
+    if(StringUtils.isNotEmpty(measureBundle.getId())) {
+      FhirContext ctx = FhirContextProvider.getFhirContext();
+      String measureContentXML = ctx.newXmlParser().encodeResourceToString(measureBundle);
+      Files.createDirectories(Paths.get(this.config.getMeasuresPath()));
+      BufferedWriter writer = Files.newBufferedWriter(
```
smailliwcs, line 50: Would recommend "try-with-resources" here to make sure the writer gets flushed/closed.  Then you could remove the explicit call to `flush` on line 54.

https://kodejava.org/how-to-write-file-using-files-newbufferedwriter/

smailliwcs (changes requested):

PR #354, opened 2022-07-26
LINK-1066

seanmcilvenna:
* `datastore` module supports basic auth and OAuth (OAuth not tested yet, but basic auth is)
* `api` module supports basic auth for data-store in a `data-store` config block but doesn't require it


amphillipsLGC:
Tested and seems good. Merging

PR #375, opened 2022-08-08
LINK-1038: Aggregate differently for each measure (or fallback on def…

arianamihailescu:
…ault)

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -8,6 +8,7 @@ import com.lantanagroup.link.api.ReportGenerator;
 import com.lantanagroup.link.auth.LinkCredentials;
 import com.lantanagroup.link.auth.OAuth2Helper;
 import com.lantanagroup.link.config.api.ApiConfigEvents;
+import com.lantanagroup.link.config.api.ApiReportDefsUrlConfig;
 import com.lantanagroup.link.config.auth.LinkOAuthConfig;
 import com.lantanagroup.link.config.query.QueryConfig;
 import com.lantanagroup.link.config.query.USCoreConfig;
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -371,10 +366,18 @@ public GenerateResponse generateReport(
 
       response.setReportId(id);
 
-      // Generate the master measure report
+      // Get the aggregator
       Class<?> reportAggregatorClass = null;
       try {
-        reportAggregatorClass = Class.forName(this.config.getReportAggregator());
+        Optional<ApiReportDefsUrlConfig> measureReportAggregatorUrl = config.getReportDefs().getUrls().stream().filter(urlConfig -> {
```
seanmcilvenna, line 372: After @smailliwcs's update to include `id` in the report def url config, this block should be altered to search for the report def url by matching the `id` of the measure report, rather than doing the url substring stuff, like we talked about on the call earlier today.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -371,10 +366,18 @@ public GenerateResponse generateReport(
...
+          String measureIdentifier = urlConfig.getUrl().substring(urlConfig.getUrl().lastIndexOf("/")+1);
+          return measureIdentifier.equalsIgnoreCase(context.getReportDefBundle().getIdentifier().getValue());
+        }).findFirst();
+        if (measureReportAggregatorUrl.isPresent()) {
+          reportAggregatorClass = Class.forName(measureReportAggregatorUrl.get().getReportAggregator());
```
seanmcilvenna, line 377: This looks great! I'm missing something, though... How comes `reportAggregator` doesn't exist in `core/src/main/java/com/lantanagroup/link/config/api/ApiReportDefsUrlConfig.java`? I would have expected a change to `ApiReportDefsUrlConfig` to add `reportAggregator` property...

seanmcilvenna (changes requested):
In general, looks good.
See my two comments. One I think requires a change. Maybe just forgot to add the change to `ApiReportDefsUrlConfig` to the PR?
Also, consider if any of this can be unit-tested. Fine to merge as-is, for now. But, maybe, if time permits, add unit tests before sprint close.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
arianamihailescu: yes

seanmcilvenna (approved):

PR #378, opened 2022-08-11
LINK-1022: Fix error handling for traditional deployments

smailliwcs:


api/src/main/java/com/lantanagroup/link/api/ApiApplication.java
```diff
@@ -64,21 +61,6 @@ public void afterPropertiesSet() throws Exception {
 
   }
 
-  /**
```
smailliwcs, line 67: Removed because it duplicates the entire Spring Security filter chain.

arianamihailescu (commented):
It looks good and is working.

Note that we might need to consolidate ErrorFilter functionality within ErrorController, cause:

1. ErrorController can be used to capture and customize what's displayed to the user, this being the class Spring provides for this purpose on both traditional and non-traditional deployments.
2. ErrorFilter should also be positioned correctly in the list of filters in order to be triggered (for example after ErrorPageFilter in traditional deployment). This order could change in the future and cause filter to not be invoked (thus causing to not have an error id associated with unhandled exceptions).

PR #381, opened 2022-08-16
LINK-1034: Use same logic to load/update a measure definition during report generation as the logic during api init

smailliwcs:


smailliwcs:
This PR includes some changes related to LINK-1022, since I locally merged that branch during development.  For review, you'll want to focus on `ApiInit` and `ReportController`.

api/src/main/java/com/lantanagroup/link/api/ApiInit.java
```diff
@@ -59,191 +58,124 @@ public class ApiInit {
...
+    // Set ID, meta, and identifiers
+    String bundleId = urlConfig.getBundleId();
+    evaluationBundle.setId(bundleId);
+    evaluationBundle.setMeta(localBundle != null
+            ? localBundle.getMeta()
+            : new Meta().addTag(Constants.MainSystem, Constants.ReportDefinitionTag, null));
+    Identifier identifier = measure.hasIdentifier()
+            ? measure.getIdentifierFirstRep()
+            : measure.addIdentifier().setSystem(Constants.MainSystem).setValue(bundleId);
+    if (!evaluationBundle.hasIdentifier()) {
```
arianamihailescu, line 168: The if can be removed to keep bundle identifier in sync with measure identifier

arianamihailescu (commented):
We have to review the use of bundle identifier versus measure identifier and the need of the bundle identifier.
Talking with Arvind bundle identifier is not generated by its script, unlike the measure identifier ( which is of course needed ).
It looks like we store that bundle identifier on the document reference but we can very well store directly the measure identifier.

arianamihailescu (approved):

api/src/main/java/com/lantanagroup/link/api/ApiInit.java
smailliwcs: Done in e9514d1.

PR #384, opened 2022-08-22
Adding public address config property to allow deployment to indicate…

seanmcilvenna:
… proxy address


amphillipsLGC (approved):
Seems okay to me.

PR #385, opened 2022-08-23
Link 978

SaraMunich-Lantana:
Created a BundlerConfig class and moved the sendWholeBundle and removeGeneratedObservations variables from ApiConfig to it. I was able to successfully generate a report and submit it with these changes.

### DIFF
--- a/nhsn/src/main/java/com/lantanagroup/link/nhsn/MeasureReportDownloader.java
+++ b/nhsn/src/main/java/com/lantanagroup/link/nhsn/MeasureReportDownloader.java
@@ -3,6 +3,7 @@ package com.lantanagroup.link.nhsn;
 import ca.uhn.fhir.context.FhirContext;
 import com.lantanagroup.link.*;
 import com.lantanagroup.link.config.api.ApiConfig;
+import com.lantanagroup.link.config.bundler.BundlerConfig;
 import org.apache.commons.io.IOUtils;
 import org.hl7.fhir.r4.model.Bundle;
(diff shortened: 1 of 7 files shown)


smailliwcs (approved):
Looks good to me.  I think we'll eventually want to directly inject `BundlerConfig` into `FhirBundler`, rather than passing it down from `ReportController` and `MeasureReportDownloader`.  But we can't do that currently, because `FhirBundler` isn't a Spring bean.  Probably better to merge this PR as is and handle the rest as part of LINK-980.

Need to make sure our configs (dev, test, etc.) get updated as part of this merge.  Probably a good idea to post whatever changes are needed to #linkdev.

nhsn/src/main/java/com/lantanagroup/link/nhsn/MeasureReportDownloader.java
```diff
@@ -22,7 +23,7 @@ public class MeasureReportDownloader implements IReportDownloader {
   protected static final Logger logger = LoggerFactory.getLogger(MeasureReportDownloader.class);
 
   @Override
-  public void download(String reportId, FhirDataProvider fhirDataProvider, HttpServletResponse response, FhirContext ctx, ApiConfig config) throws IOException, TransformerException {
```
arianamihailescu, line 25: BundleConfig could be injected in MeasureReportDownloader (Autowired) instead being passed as a parameter in download method. MeasureReportDownloader should be annotated as Component.

PR #391, opened 2022-08-26
LINK-1135: Create CLI command for ParklandInventoryImport

smailliwcs:


cli/src/main/java/com/lantanagroup/link/cli/SftpDownloader.java
```diff
@@ -18,10 +18,16 @@ public byte[] download() throws IOException, JSchException, SftpException {
...
+    try {
+      ChannelSftp channel = (ChannelSftp) session.openChannel("sftp");
+      channel.connect();
+      try (InputStream stream = channel.get(config.getPath())) {
+        return stream.readAllBytes();
+      } finally {
+        channel.exit();
+      }
+    } finally {
+      session.disconnect();
```
arianamihailescu, line 30: In finally - do we also need to disconnect the channel if is still connected?

smailliwcs: This is handled via the call to `channel.exit` above.  (`channel.disconnect`, `channel.exit`, and `channel.quit` all do the same thing.)

arianamihailescu (approved):

PR #393, opened 2022-08-29
LINK-1133: Create CLI command for KnoxMeasureReportTransfer

smailliwcs:


amphillipsLGC:
Great job putting the streams in the try block as a parameter! That is  a great way of making sure the stream closes when its finished. That often gets picked up in the security scan.

PR #394, opened 2022-09-01
Multimeasure

arianamihailescu:


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -195,6 +176,7 @@ public GenerateResponse generateReport(
           @AuthenticationPrincipal LinkCredentials user,
           HttpServletRequest request,
           @RequestParam("reportDefIdentifier") String reportDefIdentifier,
+          @RequestParam(value = "bundleIds", required = false) String[] bundleIds,
```
smailliwcs, line 179: Do we want to allow the user to specify an arbitrary set of measures?  Or should we limit this parameter to a valid multi-measure ID?  Probably something to discuss with the rest of the team.

core/src/main/java/com/lantanagroup/link/config/api/ApiMeasurePackage.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Getter
+@Setter
+@Validated
+public class ApiMeasurePackage {
+
+  @NotBlank
+  private String id;
+
+  private String[] bundleIds;
```
smailliwcs, line 17: Use `@NotNull` and `@Size(min = 1)` here to ensure that at least one bundle ID is provided?

smailliwcs (approved):

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
arianamihailescu: This is something for future - TBD

PR #395, opened 2022-09-01
Remove logic that deletes data immediately after report generation/submission

SaraMunich-Lantana:
I removed the logic from deleteSentData in ReportControler that immediately deletes data after report submission.


arianamihailescu (commented):
Do we need to remove the  delete-after-submission: true from application.yml file?

smailliwcs (approved):

smailliwcs:
Looks good to me!

PR #396, opened 2022-09-06
Link 1093

SaraMunich-Lantana:
Created a ApiDataGovernance config class that contains a RetentionPeriod class that contains the values for censusListRetention, patientDataRetention, and reportRetention.

### DIFF
--- a/api/src/main/resources/application.yml
+++ b/api/src/main/resources/application.yml
@@ -41,6 +41,11 @@ api:
   query:
     mode: Local
   measure-evaluation-threads: 10
+  data-governance:
+    retention-period:
+      census-list-retention:
+      patient-data-retention:
+      report-retention:
 thsa:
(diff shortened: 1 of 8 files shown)


arianamihailescu (changes requested):

api/src/main/resources/application.yml
```diff
@@ -71,3 +71,8 @@ uscore:
     - Location/{{locationId}}
     - Medication/{{medicationId}}
   parallel-patients: 10
+data-governance:
+  retention-period:
+    census-list-retention:
+    patient-data-retention:
+    report-retention:
```
arianamihailescu, line 78: Can we also add this section to configuration-files in SH, and to Kubernetes?

arianamihailescu (approved):

PR #401, opened 2022-09-09
LINK-1153_Update_Census_acquisition_functionality_

arianamihailescu:
…f-Interest lists for each configured Measure

### DIFF
--- a/cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListCommand.java
+++ b/cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListCommand.java
@@ -54,8 +54,8 @@ public class RefreshPatientListCommand extends BaseShellCommand {
 
   @ShellMethod(
           key = "refresh-patient-list",
-          value = "Read an Epic patient list and update the corresponding census in Link.")
-  public void execute() throws Exception {
+          value = "Read patient lists and update the corresponding census in Link.")
+  public void execute(String[] patientListPath) throws Exception {
     registerBeans();
     config = applicationContext.getBean(RefreshPatientListConfig.class);
     queryConfig = applicationContext.getBean(QueryConfig.class);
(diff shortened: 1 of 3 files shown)


cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListCommand.java
```diff
@@ -65,38 +60,23 @@ public void execute() throws Exception {
...
-    ListResource source = readList();
-    ListResource target = transformList(source, urlConfig.getCensusIdentifier());
-    updateList(target);
-  }
 
-  private ApiReportDefsUrlConfig getUrlConfig() {
-    for (ApiReportDefsUrlConfig urlConfig : apiConfig.getReportDefs().getUrls()) {
-      if (config.getPatientListId().equals(urlConfig.getPatientListId())) {
-        return urlConfig;
+    for (int i = 0; i < config.getPatientList().size(); i++) {
```
smailliwcs, line 64: Maybe not a problem, but do we want to update *every* patient list here, rather than just the ones that match a particular list ID or census identifier?

cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListCommand.java
```diff
@@ -65,38 +60,23 @@ public void execute() throws Exception {
...
+      ListResource source = readList(config.getPatientList().get(i).getPatientListId());
+      for (int j = 0; j < config.getPatientList().get(i).getCensusIdentifier().size(); j++) {
+        if (StringUtils.isNotBlank(config.getPatientList().get(i).getCensusIdentifier().get(j))) {
```
smailliwcs, line 67: Consider throwing an exception for a blank census identifier.

cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListConfig.java
```diff
@@ -6,15 +6,36 @@
...
+  private List<PatientList> patientList;
   private CensusReportingPeriods censusReportingPeriod;
 }
+
+
+@Getter
+@Setter
+class PatientList {
+  @NotBlank
+  private String patientListId;
```
smailliwcs, line 37: Recommend renaming this to `patientListPath`, since we're now expecting it to contain everything after the FHIR server base URL (at least `List/{id}`, maybe `STU3/List/{id}` for Epic on FHIR).

smailliwcs (changes requested):

cli/src/main/java/com/lantanagroup/link/cli/RefreshPatientListCommand.java
arianamihailescu: I was wondering about that too, if we need to enhance the  requirements and add a patient list id or a census identifier as a argument to create a subset and all of them if arguments not specified.

arianamihailescu: https://lantana.atlassian.net/browse/LINK-1200 captures the enhancement

PR #408, opened 2022-09-20
LINK-1158: Improve function to build Master MeasureReports for Multi-…

arianamihailescu:
…measure

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -7,6 +7,7 @@ import com.lantanagroup.link.api.ApiInit;
 import com.lantanagroup.link.api.EventController;
 import com.lantanagroup.link.api.ReportGenerator;
 import com.lantanagroup.link.auth.LinkCredentials;
+import com.lantanagroup.link.config.api.ApiMeasurePackage;
 import com.lantanagroup.link.config.api.ApiReportDefsUrlConfig;
 import com.lantanagroup.link.config.query.QueryConfig;
 import com.lantanagroup.link.config.query.USCoreConfig;
@@ -40,6 +41,7 @@ import java.text.ParseException;
 import java.util.*;
 import java.util.stream.Collectors;
 
+
 @RestController
 @RequestMapping("/api/report")
 public class ReportController extends BaseController {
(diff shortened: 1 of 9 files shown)


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -72,23 +74,13 @@ public void initBinder(WebDataBinder binder) {
...
-              reportDefIdentifier;
-
-      // Find the report def for the given identifier
-      Bundle reportDefBundle = this.getFhirDataProvider().findBundleByIdentifier(reportDefIdentifierSystem, reportDefIdentifierValue);
+    for (String bundleId : criteria.getBundleIds()) {
+      // Find the report def for the given bundleId
+      Bundle reportDefBundle = this.getFhirDataProvider().getBundleById(bundleId);
       if (reportDefBundle == null) {
-        throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Did not find report def with identifier " + reportDefIdentifier);
+        throw new ResponseStatusException(HttpStatus.NOT_FOUND, "Did not find report def with identifier " + bundleId);
```
smailliwcs, line 81: Change "identifier" to "ID" in the second argument.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -162,18 +155,64 @@ private void queryAndStorePatientData(List<String> resourceTypes, ReportContext
     }
   }
 
+
   @PostMapping("/$generate")
   public GenerateResponse generateReport(
           @AuthenticationPrincipal LinkCredentials user,
           HttpServletRequest request,
-          @RequestParam("reportDefIdentifier") String[] reportDefIdentifiers,
+          @RequestParam("bundleIds") String[] bundleIds,
```
smailliwcs, line 163: One thing to consider ... changing the `@RequestParam` argument to `bundleId` (with no *s*), but leaving the array variable name as `bundleIds` (with an *s*).  I think that reads better in the request:

```
https://.../api/report/$generate?bundleId=foo&bundleId=bar
```

Although I think Spring also allows you to send a comma-separated list as one argument, so really either way works.  Just depends on how we expect it to be used.  And if we're going to eventually encapsulate everything in a single `@RequestBody GenerateRequest` parameter at some point, it may not be worth worrying about.

arianamihailescu: Yes, comma-separated right now like: bundleIds=NHSNGlycemicControlHypoglycemicInitialPopulation,adultPatients
or bundleIds=AdultPatients%2CNHSNGlycemicControlHypoglycemicInitialPopulation%2CTHSAMeasure from UI

Yes, ideally would like to move all these parameters to a POST - I opened a tech debt ticket for it

PR #411, opened 2022-09-22
Practitioner lookup fails if name/email missing

SaraMunich-Lantana:
Added null checks to isSamePractitioner in LinkAuthenticationSuccessHandler when comparing the practitioner dynamically generated from the JWT to a previously stored practitioner for the practitioner's given name, family name, and email.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/auth/LinkAuthenticationSuccessHandler.java
+++ b/api/src/main/java/com/lantanagroup/link/api/auth/LinkAuthenticationSuccessHandler.java
@@ -16,7 +16,6 @@ import org.springframework.security.web.authentication.AuthenticationSuccessHand
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-import java.util.List;
 
 public class LinkAuthenticationSuccessHandler implements AuthenticationSuccessHandler {
   private static final Logger logger = LoggerFactory.getLogger(LinkAuthenticationSuccessHandler.class);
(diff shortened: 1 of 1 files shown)


smailliwcs (changes requested):
We can still get exceptions here.  E.g., if a practitioner is missing the `name` element entirely, then `getName().get(0)` will fail.  (HAPI will create a list of names on the fly when you call `getName`, but the list will be empty ... no element at index 0.)  Would recommend pulling this data out as follows:

```java
String familyNameX = practitionerX.getNameFirstRep().getFamily();
String givenNameX = practitionerX.getNameFirstRep().getGivenAsSingleString();
ContactPoint.ContactPointSystem telecomSystemX = practitionerX.getTelecomFirstRep().getSystem();
String telecomValueX = practitionerX.getTelecomFirstRep().getValue();
```

Each of those can still be null, though.  To compare the strings, use `StringUtils.equals(string1, string2)`, which will avoid the `NullPointerException` you might get with `string1.equals(string2)`.  `ContactPointSystem` is an enum, so you can just use `==` to compare the two values.

smailliwcs (approved):
Looks good.  FYI, HAPI has convenience functions that you can use to check for the existence of pretty much any element.  E.g., `practitioner.hasId()` will check for a non-null, non-empty ID (according to HAPI's definition of "empty" for the `IdType` class).  This works for lists, too.  E.g., `practitioner.hasName()` will first check that the `name` list itself is not null, then it will make sure that at least one item in the list is non-null and non-empty.

Just for future reference.  If you've tested this code and it's working, I don't think you need to change anything for this PR.

PR #426, opened 2022-10-11
Error in patient exclusion logic

SaraMunich-Lantana:
I got it to re-aggregate the measure report with a list of patient measure reports excluding the patients who are to be excluded.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -13,8 +13,6 @@ import com.lantanagroup.link.config.query.USCoreConfig;
 import com.lantanagroup.link.model.*;
 
 import com.lantanagroup.link.nhsn.FHIRReceiver;
-import com.lantanagroup.link.nhsn.ReportAggregator;
-
 import com.lantanagroup.link.query.IQuery;
 import com.lantanagroup.link.query.QueryFactory;
 import lombok.Setter;
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -916,19 +916,15 @@ public class ReportController extends BaseController {
 
     logger.debug("Re-evaluating measure with state of data on FHIR server");
 
-    List<MeasureReport> reportList = new ArrayList<>();
-    ListResource refs = (ListResource)measureReport.getContained().get(0);
-    for(ListResource.ListEntryComponent patient : refs.getEntry()) {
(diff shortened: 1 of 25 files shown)


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -867,25 +898,51 @@ public ReportModel excludePatients(
...
+    reportContext.getMeasureContexts().add(measureContext);
 
     logger.debug("Re-evaluating measure with state of data on FHIR server");
 
+    List<MeasureReport> reportList = new ArrayList<>();
+    ListResource refs = (ListResource)measureReport.getContained().get(0);
+    for(ListResource.ListEntryComponent patient : refs.getEntry()) {
+      for(ExcludedPatientModel excludedPatient : excludedPatients) {
+        String patientReportId = patient.getItem().getReference().substring(patient.getItem().getReference().indexOf("/") + 1);
+        String excludePatientReportId = ReportIdHelper.getPatientMeasureReportId(reportId, excludedPatient.getPatientId());
```
arianamihailescu, line 924: I think we should verify that a patientReportId is not any of the  excluded patients before adding it to the reportList.

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -867,25 +898,51 @@ public ReportModel excludePatients(
...
     // TODO: Set report def bundle on measure context
     measureContext.setMeasure(measure);
     measureContext.setReportId(measureReport.getIdElement().getIdPart());
-    context.getMeasureContexts().add(measureContext);
+    reportContext.getMeasureContexts().add(measureContext);
 
     logger.debug("Re-evaluating measure with state of data on FHIR server");
 
+    List<MeasureReport> reportList = new ArrayList<>();
+    ListResource refs = (ListResource)measureReport.getContained().get(0);
```
arianamihailescu, line 920: Should we look at all the contained lists to get all unique patients from there?

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -867,25 +898,47 @@ public ReportModel excludePatients(
...
+    try {
+      reportAggregator = (IReportAggregator) context.getBean(Class.forName(reportAggregatorClassName));
+    } catch (ClassNotFoundException e) {
+      e.printStackTrace();
+    }
     MeasureReport updatedMeasureReport = null;
+    try {
+      updatedMeasureReport = reportAggregator.generate(criteria, reportContext, measureContext);
+    } catch (ParseException e) {
+      e.printStackTrace();
```
arianamihailescu, line 940: Let exceptions to be thrown

PR #428, opened 2022-10-17
LINK-976

bstewartlg:
… codelyzer dependency due to still unresolved dependency on outdated and vulnerable angular library.


amphillipsLGC (commented):
Minor version updates shouldn't cause any problems. Glad we switched to ^ instead of ~ to reduce chances of future problems.

Looks good to me!

PR #429, opened 2022-10-18
LINK-1229:  /api/report/$generate consider passing the inputs in Requ…

arianamihailescu:
…estBody instead of RequestParams


amphillipsLGC:
Looks good to me!

PR #430, opened 2022-10-18
Link 1267 bundle ids s for generate and submit

arianamihailescu:


amphillipsLGC (commented):
Looks good, but was the check for BundleIds meant to be removed? Looks like it was added in the first commit and removed in the second. 

Line 88 in GenerateAndSubmitCommand

arianamihailescu:
> Looks good, but was the check for BundleIds meant to be removed? Looks like it was added in the first commit and removed in the second.
> 
> Line 88 in GenerateAndSubmitCommand

Yes, the validation was removed cause I am using annotations instead: 
in GenerateAndSubmitConfig file:
  @NotNull
  @Size(min = 1)
  private String[] bundleIds;

amphillipsLGC:
That makes sense, looks good!

PR #434, opened 2022-10-19
Default Date Reporting Period

SaraMunich-Lantana:
Created a default time checkbox in the UI that disables the date pickers for start date and end date and instead using Jan 1st 1900 as the default start date and the current date for the default end date.
defeault times


amphillipsLGC (approved):
Looks good to me. Only thing I see that we may want to update at some point is the text int he label for the checkbox.

PR #438, opened 2022-11-04
NHSNLink - Create a config property under api that represents the tenant ID for the facility

SaraMunich-Lantana:
Added tenantID to ApiConfig and application-dev.yml


amphillipsLGC (approved):
Looks good!

PR #439, opened 2022-11-07
MHLinkUI: changes for MHL

arianamihailescu:


amphillipsLGC (approved):
Looks good to me. Making the logo configurable was a good idea!

PR #451, opened 2022-11-15
LNK-183: Add configuration to use SAMs to validate jwt tokens

amphillipsLGC:
Made token acquisition and verification configurable to allow either general oauth token validation or SAMS token validation.


core/src/main/java/com/lantanagroup/link/config/auth/LinkOAuthConfig.java
```diff
@@ -5,16 +5,17 @@
...
 public class LinkOAuthConfig {
 
     //TODO - Add RequestType: Submitting, LoadingMeasureDef, QueryingEHR
 
     /**
-     * <strong>oauth.credential-mode</strong>
+     * <strong>oauth.credential-mode</strong> Either "client", "password", or "sams-password"
      */
-    private OAuthCredentialModes credentialMode;
+    private String credentialMode;
```
arianamihailescu, line 18: Is there a reason to remove Enumeration and made it String?

core/src/main/java/com/lantanagroup/link/config/auth/LinkOAuthConfig.java
```diff
@@ -44,16 +50,23 @@ public class LinkOAuthConfig {
     public boolean hasCredentialProperties() {
         if (this.credentialMode != null) {
             switch (this.credentialMode) {
-                case Client:
+                case "clilent":
```
arianamihailescu, line 53: Looks like a type in the client word

core/src/main/java/com/lantanagroup/link/auth/OAuth2Helper.java
```diff
@@ -450,22 +513,31 @@ public static String getToken(LinkOAuthConfig authConfig, CloseableHttpClient cl
       logger.info("Configured to authentication when submitting. Requesting a token from configured token URL");
 
       switch (authConfig.getCredentialMode()) {
-        case Client:
+        case "client":
           token = OAuth2Helper.getClientCredentialsToken(
                   client,
                   authConfig.getTokenUrl(),
-                  authConfig.getUsername(),
+                  authConfig.getClientId(),
```
arianamihailescu, line 520: In LinkOAuthConfig looks like for "Client" credential mode,  credentials properties checked are username and password, but here seems to be client id and password ?

core/src/main/java/com/lantanagroup/link/auth/SamsTokenValidator.java
```diff
@@ -0,0 +1,73 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+
+@Component
+public class SamsTokenValidator implements  ITokenValidator {
+  private static final Logger logger = LoggerFactory.getLogger(SamsTokenValidator.class);
+  String tokenValidationUrl = ""; //get from config
```
arianamihailescu, line 21: Will this tokenValidationUrl  be added later and used?

core/src/main/java/com/lantanagroup/link/config/auth/LinkOAuthConfig.java
amphillipsLGC: Ooops! Fixed now, good catch!

core/src/main/java/com/lantanagroup/link/auth/OAuth2Helper.java
amphillipsLGC: Good catch, it should have checked for clientId instead of username, fixed!

core/src/main/java/com/lantanagroup/link/auth/SamsTokenValidator.java
amphillipsLGC: Yeah that is probably a good idea!  Updated Api and DataStore configs to include tokenValidationEndpoint, updated LinkAuthManager to accept a parameter for the token validation endpoint.

PR #480, opened 2023-01-10
LINK-290: added new group to count the total number of records

arianamihailescu:


mhl/src/main/java/com/lantanagroup/link/mhl/MHLAggregator.java
```diff
@@ -78,6 +80,14 @@ public void aggregatePatientReports(MeasureReport masterMeasureReport, List<Meas
...
           childrenPopulation.getCode().addCoding(childrenCoding);
           childrenPopulation.setCount(population.getCount() == 0 ? 0 : storeSupplementalData(measureReport));
+
+          Coding totalRecordsCoding = new Coding()
+                  .setSystem(TOTAL_NUMBER_OF_RECORDS_QUERIED_SYSTEM)
+                  .setCode(TOTAL_NUMBER_OF_RECORDS_QUERIED_CODE)
+                  .setDisplay("Total number of records queried");
+          MeasureReport.MeasureReportGroupPopulationComponent totalPopulation = group.addPopulation();
+          totalPopulation.getCode().addCoding(totalRecordsCoding);
+          totalPopulation.setCount(1);
```
smailliwcs, line 90: I assume we always `setCount(1)` because it represents the mother?  And then the aggregator takes care of summing them?

smailliwcs (approved):

mhl/src/main/java/com/lantanagroup/link/mhl/MHLAggregator.java
arianamihailescu: That's right

PR #509, opened 2023-05-08
LNK-579: Adding version info to reported measure bundle

ajagann1:


arianamihailescu (approved):
Looks good to me

arianamihailescu (approved):
Looks good to me

PR #518, opened 2023-06-15
LNK-726: Added in some basic checks to only copy over both a value an…

ajagann1:
…d a system of an identifier to the type (if both don't exist, it simply doesn't).  Also made it so that no duplicates get generated from the copy/paste


amphillipsLGC (approved):
Looks good, thanks!

PR #529, opened 2023-06-28
SharedService SQL Server Conversion

MontaltoNick:
Remaining items to convert SharedService.java to use SQL Server in place of Mongo.

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/db/SharedService.java
+++ b/core/src/main/java/com/lantanagroup/link/db/SharedService.java
@@ -161,7 +161,7 @@ public class SharedService {
 
       if(rs.next()) {
         var json = rs.getString(1);
-        java.util.Date lastUpdated  = rs.getTimestamp(2);
+        java.util.Date lastUpdated  = new java.util.Date(rs.getTimestamp(2).getTime());
         measureDef.setBundle(FhirContextProvider.getFhirContext().newJsonParser().parseResource(Bundle.class, json));
         measureDef.setMeasureId(measureId);
         measureDef.setLastUpdated(lastUpdated);
--- a/core/src/main/java/com/lantanagroup/link/db/SharedService.java
+++ b/core/src/main/java/com/lantanagroup/link/db/SharedService.java
@@ -3,6 +3,7 @@ package com.lantanagroup.link.db;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.lantanagroup.link.FhirContextProvider;
 import com.lantanagroup.link.Hasher;
 import com.lantanagroup.link.auth.LinkCredentials;
 import com.lantanagroup.link.config.MongoConfig;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -176,69 +164,106 @@ public ArrayList<Tenant> getTenantConfigs() {
...
+      throw new RuntimeException(e);
+    } catch (JsonMappingException e) {
+      throw new RuntimeException(e);
+    } catch (JsonProcessingException e) {
+      throw new RuntimeException(e);
     }
+  }
 
-    return this.client;
+  public ArrayList<MeasureDefinition> getMeasureDefinitions() {
```
seanmcilvenna, line 234: Prefer to respond with the `List` interface

core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -251,13 +276,14 @@ public int deleteTenantConfig(String tenantId) {
...
+  public void saveMeasureDefinition(MeasureDefinition measureDefinition) {
     try (Connection conn = this.getSQLConnection()) {
       assert conn != null;
 
-      SQLCSHelper cs = new SQLCSHelper(conn, "{ CALL saveTenant (?, ?) }");
-      cs.setNString("tenantId", tenant.getId());
-      cs.setNString("json", mapper.writeValueAsString(tenant));
+      SQLCSHelper cs = new SQLCSHelper(conn, "{ CALL saveMeasureDef (?, ?, ?) }");
+      cs.setNString("measureId", measureDefinition.getMeasureId());
+      cs.setNString("bundle", mapper.writeValueAsString(measureDefinition.getBundle()));
```
seanmcilvenna, line 285: Since `bundle` represents FHIR, I suggest using the HAPI FHIR serializer/deserializer. It handles a number of things about the FHIR spec that the default serializers do not.
Example:
`FhirContextProvider.getFhirContext().newJsonParser().resourceAsString(measureDefinition.getBundle());`
and to deserialize:
`FhirContextProvider.getFhirContext().newJsonParser().parseResource(Bundle.class, jsonString);`

core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -270,28 +296,193 @@ public void saveTenantConfig(Tenant tenant){
...
+      logger.debug("X-REAL-IP is: " + request.getHeader("X-REAL-IP") + " and is being used for remoteAddress");
+      remoteAddress = request.getHeader("X-REAL-IP");
+    } else {
+      logger.debug("X-REAL-IP IP is not found.");
+      remoteAddress = request.getRemoteAddr() != null ? (request.getRemoteHost() != null ? request.getRemoteAddr() + "(" + request.getRemoteHost() + ")" : request.getRemoteAddr()) : "";
+    }
+    return remoteAddress;
+  }
+
+  public MongoClient getClient() {
```
seanmcilvenna, line 463: Does this class need `getClient(): MongoClient` anymore?

seanmcilvenna (changes requested):

core/src/main/java/com/lantanagroup/link/db/SharedService.java
MontaltoNick: Okay - do I only need this here and the other serialize operations are fine?

core/src/main/java/com/lantanagroup/link/db/SharedService.java
MontaltoNick: I will double check and remove if not.

MontaltoNick: The 'public MongoClient getClient()' function is used in the TenantService (that Steven is working on. this reference might go away) and the BulkStatusService. Is the BulkStatusService something that we will be keeping around and need to adapt as well? Or will that be ripped out with the conversion?

core/src/main/java/com/lantanagroup/link/db/SharedService.java
MontaltoNick: I converted this as requested.

core/src/main/java/com/lantanagroup/link/db/SharedService.java
MontaltoNick: Done!

core/src/main/java/com/lantanagroup/link/db/SharedService.java
MontaltoNick: It looks like Steven deleted this in his branch - so this should be sorted out once we merge everything together.

PR #555, opened 2023-09-22
 WIP Direct Measure eval integration

JPercival:
- Rebased onto master
- Fixed some dependencies
- Work towards fixing tests
- Test runs, but fails due to some missing content


smailliwcs:
Addressed in #701.

PR #556, opened 2023-09-27
LNK-1336: Finalizing validation changes

seanmcilvenna:
* Updates to unit tests. Not all unit tests are based on the large patient bundle.
* Updated the patient bundles in `test/resources` with the latest report generation format
* Moved away from using HAPI validation libraries, and now directly using FHIR core validation libraries
* Extracting specific profiles and code systems that are needed for validation, not just simply loading all packages/tgz filse (which is inefficient and bloated with profiles/valuesets/codesystems we don't need)
* Removed configuration of validation. It always loads all the profiles regardless of configuration, and validates against whatever profiles are asserted in the data/resources
* Updated the `Validator` class to run as a singleton in the Spring Boot application, so that it only initializes once (which takes some time)


api/src/main/resources/build.yml
smailliwcs, line 1: Don't think this file belongs in the PR.  Or if it does, it needs to be updated to refer to a more recent commit.  But on second thought, do we want this file committed to the repo at all?  As soon as you commit it, the `commit` property is out of date ... you can only ever refer to the *parent* commit.  So is the intention that this file be supplied at deploy time?  And continuing the tangent, shouldn't we be reading the CQF Ruler version *from* CQF Ruler?  We won't necessarily know it at deploy time, and if we happen to upgrade CQF Ruler after initial deployment, we'll have to remember to go back and update `build.yml`.

core/src/main/resources/cqfmeasures.tgz
smailliwcs, line 1: Is there a technical reason we can't use these NPM packages as tarballs?  I.e., are we *required* to extract the individual conformance resources into the repo?  Because I think that will get unwieldy.  Ideally we'd use something like this:

```java
private void addResources(org.springframework.core.io.Resource npmPackageResource) {
    NpmPackage npmPackage;
    try (InputStream stream = npmPackageResource.getInputStream()) {
        npmPackage = NpmPackage.fromPackage(stream);
    }
    for (String fileName : npmPackage.listResources("StructureDefinition", "CodeSystem", "ValueSet")) {
        IBaseResource fhirResource;
        try (InputStream stream = npmPackage.loadResource(fileName)) {
            fhirResource = fhirContext.newJsonParser().parseResource(stream);
        }
        prePopulatedValidationSupport.addResource(fhirResource);
    }
}
```

smailliwcs: And if we *can't* use the tarballs, do we still need `terminology.tgz`, which remains in `core`'s resources?

core/src/main/java/com/lantanagroup/link/validation/Validator.java
```diff
@@ -1,80 +1,254 @@
...
+      case NULL:
+        return OperationOutcome.IssueSeverity.NULL;
+      default:
+        throw new RuntimeException("Unexpected severity " + severity);
+    }
+  }
+
+  private void validateResource(Resource resource, OperationOutcome outcome, OperationOutcome.IssueSeverity severity, Integer entryIndex) {
+    ValidationOptions opts = new ValidationOptions();
+    resource.getMeta().getProfile().forEach(p -> opts.addProfile(p.getValue()));
```
smailliwcs, line 177: Doesn't hurt anything, but this is unnecessary.  The instance validator already validates against declared profiles.  Explicitly adding profiles to `ValidationOptions` is only necessary if you want to validate against profiles that *aren't* declared.

api/src/main/resources/build.yml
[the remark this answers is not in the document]
seanmcilvenna: You're totally right. It was an accidental additional. I'll remove.

core/src/main/java/com/lantanagroup/link/validation/ResourceFetcher.java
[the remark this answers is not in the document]
seanmcilvenna: I'm actually not entirely sure _why_ this ResourceFetcher is needed. I just know that I saw it in the HAPI validation implementation and it seemed to be required (through a null exception if it was not provided). I added all these logs in each method to find out which of the methods validation was using, and found it was only using the `resolveUrl()` method. It seems that it uses `resolveUrl()` to do a pre-check on what URLs/resources the validator has available to it. If `resolveUrl()` returns false, then the validator includes a "Cannot resolve url XX" as an error. I'll include the above comments as a class-level comment.

core/src/main/resources/cqfmeasures.tgz
seanmcilvenna: And no, removing `terminology.tgz`

core/src/main/java/com/lantanagroup/link/validation/Validator.java
seanmcilvenna: Huh! I thought I had observed different behavior in my tests. I removed it, and re-ran the tests, and you're right. Oddly, though, one of my tests is returning an extra error regardless of whether or not this line exists or not. Not sure how my unit tests worked when I submitted the PR, and don't now. But, I'm fixing and committing.

MontaltoNick (approved):

PR #558, opened 2023-10-05
Data Tracing

smailliwcs:


core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -61,31 +60,11 @@ public Connection getSQLConnection() {
...
-          logger.error("Failed to execute statement to initialize shared db", e);
-          return;
-        }
-      }
-    } catch (SQLException | NullPointerException e) {
-      logger.error("Failed to connect to shared database", e);
-    } catch (IOException e) {
-      logger.error("Could not read shared-db.sql file", e);
+    try (Connection connection = this.getSQLConnection()) {
+      SQLScriptExecutor.execute(connection, new ClassPathResource("shared-db.sql"));
```
seanmcilvenna, line 64: 👍 I thought about doing this... Thanks for taking the time _to_ do it :)

core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -578,6 +557,17 @@ public void saveUser(User user) {
     }
   }
 
+  private User createUser(ResultSet rs) throws SQLException {
```
seanmcilvenna, line 560: I liked the idea of having this simple logic in the model class... But, I can also see the value in keeping the stuff related to `ResultSet` in the service. Probably a better call. 👍

query/src/main/java/com/lantanagroup/link/query/uscore/BulkQuery.java
```diff
@@ -90,13 +89,14 @@ public void executeInitiateRequest(TenantService tenantService, BulkStatusServic
...
+//      bulkStatus.setErrorMessage(sbuilder.toString());
+      tenantService.saveBulkStatus(bulkStatus);
+      logger.warn(sbuilder.toString());
     } catch (InterruptedException e) {
       logger.warn("Interrupted while initiating bulk export", e);
-      bulkStatus.setStatus(BulkStatuses.pending);
-      service.saveBulkStatus(bulkStatus);
+      bulkStatus.setStatus(BulkStatuses.PENDING);
+      tenantService.saveBulkStatus(bulkStatus);
       Thread.currentThread().interrupt();
```
seanmcilvenna, line 100: If we're catching an interrupt, why cause a new interrupt? Why not just log an error/warning and just `return`?

query/src/main/java/com/lantanagroup/link/query/uscore/QuerySaver.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.Date;
+import java.util.List;
+import java.util.UUID;
+
+@Interceptor
+public class QuerySaver {
```
seanmcilvenna, line 19: Cool use of an interceptor for this

seanmcilvenna (approved):

core/src/main/java/com/lantanagroup/link/db/SharedService.java
smailliwcs: Yeah, shared and tenant patterns have been quite different from the get-go.  Ideally they'd be more consistent, but I tried to keep the refactors in this PR mostly limited to tenant-specific stuff.  Someday ...

query/src/main/java/com/lantanagroup/link/query/uscore/BulkQuery.java
smailliwcs: Not sure.  All the bulk-data-related changes should have been stylistic/structural, so would need to defer to Ed on the intention here.

query/src/main/java/com/lantanagroup/link/query/uscore/QuerySaver.java
smailliwcs: 🙂 Spent a while staring at the screen before hitting on this strategy.  Still a little spooky because it relies on the `X-Request-ID` header (conceptually, the `query.id`) having been set elsewhere (e.g., in `PatientData`).  But that's necessary because we need the ID in both places ... here in the interceptor (where we have access to the request URL and body) so we can persist the `query`, and at the query invocation site (where we have access to the response _as FHIR resources_) so we can persist the `dataTrace`s.  So an acceptable level of spookiness, I think, given how much we'd have needed to tear up the `query` module otherwise.

PR #564, opened 2023-11-01
Store logs in DB

seanmcilvenna:
Updated shared-db.sql to create the logback sql server tables (three of them)
Logback configuration changes to use DBAppender to store logs in the three tables

### DIFF
--- a/api/src/main/resources/logback.xml
+++ b/api/src/main/resources/logback.xml
@@ -13,8 +13,10 @@
         </connectionSource>
     </appender>
 
-    <logger name="com.lantanagroup.link" level="DEBUG">
-        <appender-ref ref="link-db"/>
-    </logger>
+    <logger name="com.lantanagroup.link" level="DEBUG"/>
     <logger name="org.springframework.web" level="DEBUG"/>
+
+    <root>
+        <appender-ref ref="link-db"/>
+    </root>
 </configuration>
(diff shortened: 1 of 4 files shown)


api/src/main/resources/logback.xml
```diff
@@ -3,7 +3,18 @@
...
+    <appender name="link-db" class="ch.qos.logback.classic.db.DBAppender">
+        <connectionSource class="ch.qos.logback.core.db.DriverManagerConnectionSource">
+            <driverClass>com.microsoft.sqlserver.jdbc.SQLServerDriver</driverClass>
+            <url>${dbUrl}</url>
+        </connectionSource>
+    </appender>
+
+    <logger name="com.lantanagroup.link" level="DEBUG">
+        <appender-ref ref="link-db"/>
+    </logger>
```
smailliwcs, line 18: Is this intentional?  We only want to see Link's logs in the database?  Wondering if the database appender should be attached to the root logger instead, so that we see `INFO`/`WARN`/`ERROR` messages from other loggers as well.

core/pom.xml
```diff
@@ -67,6 +67,11 @@
             <artifactId>hapi-fhir-client</artifactId>
             <version>${hapi.lib.version}</version>
         </dependency>
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
+            <version>1.2.6</version>
+        </dependency>
```
smailliwcs, line 74: Not strictly necessary ... this should already appear on the classpath thanks to the Spring Boot starter.

smailliwcs (approved):

PR #567, opened 2023-11-06
Addl functionality for deletes

edward-miller-lcg:
Add to delete all reports and patient lists when deleting by report id or patient list id.


MontaltoNick (approved):
Looks Good!

PR #568, opened 2023-11-10
LNK-1406: Change Query Plan configuration to skip supplemental data check when specified

smailliwcs:
In some cases, we may only need an initial query/evaluate phase, without a subsequent supplemental phase.  This commit updates `ReportController` to recognize such a case, suppress the supplemental phase, and skip straight to aggregation.


amphillipsLGC (approved):
Looks good!

PR #569, opened 2023-11-10
LNK-1246: QA needs conditional API endpoints for DELETE functionality

smailliwcs:


amphillipsLGC (approved):
Looks good! I like this format of SQL statements and parameters better than some of the other stuff in the shared service. Might be worth updating other methods to follow this format. I find the  many "?" characters to be confusing where this is more easy to understand.

PR #572, opened 2023-11-14
LNK-1246: QA needs conditional API endpoints for DELETE functionality

smailliwcs:


seanmcilvenna:
Based on yesterday's conversation, it would be good to have the swagger.yml spec updated for this QA controller/endpoints... It may be too much work/time to include in this PR, though.

edward-miller-lcg (approved):

PR #575, opened 2023-11-15
Logging to DB after DB is initialized

seanmcilvenna:


core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -59,6 +62,26 @@ public Connection getSQLConnection() {
...
+
+    DBAppender appender = new DBAppender();
+    appender.setContext(logCtx);
+    appender.setConnectionSource(source);
+    appender.setName("link-db");
+    appender.start();
+
+    logCtx.getLogger("ROOT").addAppender(appender);
+
+    logger.warn("test");
```
smailliwcs, line 82: Assume this test message can be removed?

pom.xml
```diff
@@ -65,7 +65,7 @@
         <dependency>
             <groupId>com.microsoft.sqlserver</groupId>
             <artifactId>mssql-jdbc</artifactId>
-            <version>12.4.2.jre11</version>
+            <version>12</version>
```
smailliwcs, line 68: This causes a dependency resolution error for me.  Are you able to `mvn clean compile` with this in place?

smailliwcs (changes requested):
FWIW, in my testing, Logback seemed robust enough to keep plugging away despite a nonexistent database, then start using it once `shared-db.sql` was applied.  But definitely preferable this way, if for no other reason than to reduce the log noise associated with the errors that would get thrown.

pom.xml
smailliwcs: Looks like the CI pipeline also failed because of this.

seanmcilvenna: This was definitely a mistake. Didn't mean for this. I'm rolling back the change now.

core/src/main/java/com/lantanagroup/link/db/SharedService.java
seanmcilvenna: Yup, removed.

smailliwcs (approved):

PR #576, opened 2023-11-18
Validation enhancements

seanmcilvenna:
* Bundling is setup to sort the entries of the bundle the same way, every time (FYI Steven - see FhirBundleEntrySorter class, and there's a basic unit test for this)
* The "Device" attached to the submission bundle and the validation results that represents the state of NHSNLink is created at the time of report generation, instead of at the time of submission, and persisted in the "report" table. This way, the Device accurately represents NHSNLink at the time the report was generated. The persisted device information is pulled out from the "report" table and included in the submission when we get to that point, as well as when validation results are returned as an OperationOutcome.
* Validation is automatically executed as part of report generation (at the very end).
  * This may cause generation to take a bit longer
  *This is necessary so that we can persist the validation results and make them available to our UI, which is hopefully around the corner
* The endpoints for validation have changed and look roughly like this (more consistent):
  * POST /api/validate (post a bundle in the request body to see how validation behaves against it - for testing purposes)
  * POST /api/validate/summary (same as above, but a simple de-duplicated summary of messages so they can be reviewed "at a glance")
  * GET /api/validate/:tenantId/:reportId (get the persisted validation results for a given tenant and report)
  * GET /api/validate/:tenantId/:reportId/summary (get the summary of persisted validation results for a given tenant and report)
  * POST /api/validate/:tenantId/:reportId (re-validate an already generated submission bundle and persist the results)
* The endpoints are updated in Postman in the Validation > "Pending Changes" folder
* The swagger specification at api-docs.nhsnlink.org is also updated, and accurate, according to these changes
* I've made some improvements the actual process of validation as well, such as:
  * Added some missing code systems (Rob is in parallel finding some of the other missing code systems, and we can add those in when he's done)
  * Updated how URLs in the bundles are resolved, which was causing issues for extension urls such as "ombCategory" and "details"
* The GET endpoints above for retrieving validation results returns an OperationOutcome instead of a Bundle
* I also found a "works on my computer" situation that I resolved. A lot of the profiles embedded into NHSNLink were not being loaded in the deployed/DEV environment. So I was getting a lot of errors in DEV related to "cannot resolved link-received-date profile" (for example). This was an issue with deployment, and seems to be resolved.
* Per discussions with Corey/Dave, changed the query plan that's embedded in the submission bundle from a Binary resource to a Library resource that has some identifying information in it (such as a "type" and a "name")


seanmcilvenna:
Hold off on merging this... Over the last week I added additional changes to support validation message categories re LNK-1385. I just have a little bit to wrap up on the changes, then I'll update the PR.

seanmcilvenna:
@smailliwcs this is ready for final review. I know it's a big PR. Let me know if you'd like to review on a call together.

smailliwcs (approved):
Gonna approve this so we can get it out on dev, because I think there's some ongoing confusion about whether these changes are live.  (They're not ... Nick made an intervening merge to `development` after the manual push of `val_enhancements` to dev.)  But if you have a few minutes to meet, I'll take you up on the offer.  Would like to understand these changes a bit better.

api/pom.xml
smailliwcs, line 1: Was explicitly adding `maven-war-plugin` here necessary to fix the deployment issue?  Because based on the parent (Spring Boot) POMs, we should already have this version available (implicitly).

core/src/main/java/com/lantanagroup/link/FhirBundler.java
```diff
@@ -97,9 +107,12 @@ private Binary getQueryPlanBinary(List<String> measureIds) {
 
     Yaml yaml = new Yaml();
     String queryPlansYaml = yaml.dump(queryPlans);
-    binary.setContentAsBase64(Base64.getEncoder().encodeToString(queryPlansYaml.getBytes()));
+    String queryPlansBase64 = Base64.getEncoder().encodeToString(queryPlansYaml.getBytes());
+    lib.addContent()
+            .setContentType("text/yml")
+            .setData(queryPlansBase64.getBytes());
```
smailliwcs, line 113: When converting strings to byte arrays, I think we'll want to specify UTF8.  Otherwise we'll get the system's default charset.

core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
```diff
@@ -0,0 +1,262 @@
...
+
+  private static List<String> getPatientIds(Bundle bundle) {
+    return bundle.getEntry().stream()
+            .filter(e -> e.getResource().getResourceType().equals(ResourceType.Patient))
+            .map(e -> e.getResource().getIdElement().getIdPart())
+            .collect(Collectors.toList());
+  }
+
+  private static Bundle.BundleEntryComponent getLinkOrganization(Bundle bundle) {
+    // TODO: Need a better way to identify the link organization
```
smailliwcs, line 23: Look for `Constants.SubmittingOrganizationProfile`?

core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
```diff
@@ -0,0 +1,262 @@
...
+    return bundle.getEntry().stream()
+            .filter(e -> e.getResource().getResourceType().equals(ResourceType.Organization))
+            .findFirst()
+            .orElse(null);
+  }
+
+  private static Bundle.BundleEntryComponent getLinkDevice(Bundle bundle) {
+    // TODO: Need a better way to identify the link device
```
smailliwcs, line 31: No profile to look for here, and this one's probably the bigger problem (compared to `getLinkOrganization`), because whereas we never pull `Organization`s from the EHR (so they'll never appear in patient data, at least for the measures we're currently running), we _do_ pull `Device`s for ACH.

core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
```diff
@@ -0,0 +1,262 @@
...
+   * @param patientId The patient id
+   * @return True if the resource is related to the patient
+   */
+  private static boolean isResourceRelatedToPatient(Resource resource, String patientId) {
+    switch (resource.getResourceType()) {
+      case Patient:
+        return resource.getIdElement().getIdPart().equals(patientId);
+      case Encounter:
+        Encounter encounter = (Encounter) resource;
+        return encounter.getSubject().hasReference() && encounter.getSubject().getReference().equals("Patient/" + patientId);
```
smailliwcs, line 158: Recommend defining a `boolean isReferenceToPatient(Reference)` helper method that encapsulates these two tests.

core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -260,21 +258,21 @@ public static Device getDevice(ApiConfig apiConfig) {
 
     if (StringUtils.isNotEmpty(apiVersionModel.getBuild())) {
       device.addVersion()
-              .setType(new CodeableConcept().addCoding(new Coding().setCode("build")))
+              .setType(new CodeableConcept().addCoding(new Coding().setCode("build").setSystem("http://www.cdc.gov/nhsn/fhirportal/dqm/ig/CodeSystem/codesystem-link-device-version")))
```
smailliwcs, line 261: Need a new `Constants` field for this code system?

smailliwcs (approved):

core/src/main/java/com/lantanagroup/link/FhirHelper.java
seanmcilvenna: Added `Constants.LinkDeviceCodeSystem`

core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
seanmcilvenna: Actually, we _do_ have a profile for this. I've modified `getLinkDevice()` to look for it.

core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
seanmcilvenna: Updated to match with that constant/profile.

api/pom.xml
seanmcilvenna: Removed this dependency.

PR #586, opened 2023-12-13
test pr to see if bundle size error goes away

fjorge-saran:


fjorge-saran:
This was just to verify if we get any bundle errors.

PR #594, opened 2023-12-14
LNK-1816 - configurable to apply db schema changes or not

seanmcilvenna:


api/src/main/java/com/lantanagroup/link/api/ApiInit.java
```diff
@@ -128,7 +133,12 @@ public void init() {
       return;
     }
 
-    this.initDatabases(tenants);
+    if (!this.config.isApplySchemas()) {
+      logger.warn("Not configured to apply schemas to database. Skipping tenant database schema init.");
+      return;
```
smailliwcs, line 138: Was this early return intentional?  Wouldn't we still want to perform the remaining checks in this method?

smailliwcs (changes requested):

api/src/main/java/com/lantanagroup/link/api/ApiInit.java
seanmcilvenna: Removed the `return`. Good catch. Not sure what I was thinking

smailliwcs (approved):

PR #610, opened 2023-12-21
feature/SLNL-72_Api_calls_for_facilities

fjorge-saran:
- added facilities-api.service file to handle api calls for facilities and loaded facilities table data.
- moved all auth files to services/auth folder
- moved dataService file under services/api folder
- created services/api/facilities folder to move all common api handlers for facilities, This is the format I will use to create another file.


szydlofj (approved):
Nice work

PR #611, opened 2023-12-21
Validation optional as part of $generate

seanmcilvenna:


smailliwcs (approved):
Approving and merging, but we should keep in mind that, when `GET /api/validate/{tenant-id}/{report-id}` responds with a clean report, it's now ambiguous whether validation was skipped or there were truly no issues.

seanmcilvenna:
Great point. Probably would be good to add an "INFORMATION" validation message that asserts "validation was not performed on this report". I may do that sometime over the next couple weeks...

PR #612, opened 2023-12-21
created a link component to use throughout the site, removed resource…

szydlofj:
…s-contents component since it is page specific

Cleanup task as I was looking at the style guide. I think it makes sense to build these cards directly onto the page rather than creating a separate component for the specific pages content?

Created a link component that automatically adds the external icon and target '_blank' if an external link


fjorge-saran (approved):
Thanks for removing the resource-contents component. I does seem to be obsolete.

PR #644, opened 2024-01-11
 LNK-1861 Performance improvements to sorting

seanmcilvenna:
Temporary unit test against a large bundle came down from 800ms to 200ms.


smailliwcs (approved):
Made some minor tweaks, but looks good!

PR #645, opened 2024-01-11
LNK-1861 Performance improvements to sorting for RC_0.3.2

seanmcilvenna:
Temporary unit test against a large bundle came down from 800ms to 200ms.


smailliwcs (approved):
Approved after cherry-picking my tweaks from `LNK-1861`.

PR #665, opened 2024-01-23
LNK-1874 Deduplicate Multi Measure

MontaltoNick:


smailliwcs (approved):
Good call.  Should resolve the duplication that this class is responsible for.  And it sounds like the issue of legit duplicates getting promoted from separate measure reports is going to be handled in a separate ticket.

PR #673, opened 2024-01-25
Misc. Testing Improvements

seanmcilvenna:
Adding ability to see total number of patients in each patient list of the patient identifier controller
Adding ability to bypass querying directly int he $generate and $generateMultiMeasure endpoints, rather than solely relying on system config
Added a progress report logging message for evaluation

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
+++ b/api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
@@ -21,8 +21,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.text.ParseException;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.ForkJoinPool;
+import java.util.List;
+import java.util.concurrent.*;
 
 /**
  * This class creates a master measure report based on every individual report generated for each patient included in the "census" list..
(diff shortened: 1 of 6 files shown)


api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -76,9 +79,22 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
...
+      }));
+
+      ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+      scheduler.scheduleAtFixedRate(() -> {
+        long remainingTasks = latch.getCount();
+        long totalTasks = pois.size();
+        double completionPercentage = ((totalTasks - remainingTasks) / (double) totalTasks) * 100;
+
+        logger.info("Progress of measure evaluation for report {} count: {}, Completion: {}%", report.getId(), remainingTasks, completionPercentage);
+      }, 0, 5, TimeUnit.SECONDS);
```
smailliwcs, line 94: Worried about the complexity here.  Any reason not to use an `AtomicInteger` instead of a latch?  Could decrement and report its value (relative to `pois.size()`) after every evaluation.

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -76,9 +79,22 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
...
+
+      latch.await();
```
smailliwcs, line 96: Do we need to catch `InterruptedException` here?

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -76,9 +79,22 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
...
+      scheduler.shutdown();
```
smailliwcs, line 97: Does this need to happen in a `finally` block?

smailliwcs (approved):

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -76,6 +81,10 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
                   }
                 } catch (Exception e) {
                   logger.error("Error generating measure report for patient {}", patient.getId(), e);
+                } finally {
+                  int completed = progress.incrementAndGet();
+                  double percent = Math.round((completed * 100.0) / measureContext.getPatientsOfInterest(queryPhase).size());
+                  logger.info("Progress ({}%) for report {} is {} of {}", String.format("%.2f", percent), reportContext.getMasterIdentifierValue(), progress.get(), measureContext.getPatientsOfInterest(queryPhase).size());
```
smailliwcs, line 87: In this block, replace `measureContext.getPatientsOfInterest(queryPhase)` with `pois` to avoid recalculation.  Should also replace `progress.get()` with `completed`, since another thread may change the value of `progress` between lines 85 and 87, while `completed` reflects its value as atomically incremented in *this* thread.

PR #674, opened 2024-01-25
LNK-1825

MontaltoNick:
Add Category handling to metrics output logging.

https://lantana.atlassian.net/browse/LNK-1825


core/src/main/java/com/lantanagroup/link/time/StopwatchManager.java
```diff
@@ -37,14 +35,24 @@ public synchronized String getStatistics() {
             .mapToInt(String::length)
             .max()
             .orElse(DURATION_WIDTH);
-    Formatter formatter = new Formatter(taskWidth, COUNT_WIDTH, DURATION_WIDTH);
+
+    var maxCategory = durationsByTask.entrySet().stream().max((s,e) -> s.getKey().substring(s.getKey().indexOf(":") + 1).length()).get();
+    var categoryWidth = maxCategory.getKey().substring(maxCategory.getKey().indexOf(":") + 1).length() + 5;
```
smailliwcs, line 40: Not following this logic.  The lambda supplied to `Stream.max` is a `Comparator`:

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Stream.html#max(java.util.Comparator)

Don't think that's what you want here.  Would recommend mapping to an `IntStream` first and *then* calling `max`.  Should also update the calculation of `taskWidth` above.  Something like this:

```java
int taskWidth = durationsByTask.keySet().stream()
        .mapToInt(key -> key.indexOf(":"))
        .max()
        .orElse(0);
taskWidth = Integer.max(taskWidth, TASK_WIDTH_MIN);
int categoryWidth = durationsByTask.keySet().stream()
        .mapToInt(key -> key.length() - key.indexOf(":") - 1)
        .max()
        .orElse(0);
categoryWidth = Integer.max(categoryWidth, CATEGORY_WIDTH_MIN);
```

Note that there's no `TASK_WIDTH_MIN` currently defined, but since you added one for the category, we should probably do the same for task.

core/src/main/java/com/lantanagroup/link/time/StopwatchManager.java
```diff
@@ -37,14 +35,24 @@ public synchronized String getStatistics() {
...
+
+    if(categoryWidth < CATEGORY_WIDTH_MIN){
+      categoryWidth = CATEGORY_WIDTH_MIN;
+    }
```
smailliwcs, line 44: Remove if using the `Integer.max` call suggested above.

smailliwcs (changes requested):

core/src/main/java/com/lantanagroup/link/time/StopwatchManager.java
MontaltoNick: My goal was to get the category with the largest value to use as the column width. I wasn't aware of the mapToInt functionality in Java - that is more concise compared to what I was doing.

smailliwcs (approved):

PR #684, opened 2024-01-30
LNK-1844 & LNK-1846

MontaltoNick:
Validation Issue and Report Generation Metrics


api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
```diff
@@ -94,14 +94,14 @@ public MetricsReportResponse getMetricReport(
...
-    MetricsReportResponse report = this.calculatePeriodMetrics(period, endDate, metrics);
+    MetricsReportResponse report = this.calculatePeriodMetrics(period, endDate, metrics, tenantId);
 
     return report;
   }
 
   //TODO: This functionality probably should be broken out into a service
 
-  private MetricsReportResponse calculatePeriodMetrics(String period, LocalDate end, List<Metrics> metrics) {
+  private MetricsReportResponse calculatePeriodMetrics(String period, LocalDate end, List<Metrics> metrics, String tenantId) {
```
smailliwcs, line 104: The addition of `tenantId` here appears to be unnecessary (it's unused within `calculatePeriodMetrics`).  Remove here and in the above invocation?

api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
```diff
@@ -239,7 +250,7 @@ private double calculateQueryTimeAvg(List<Metrics> periodMetrics, List<String> r
 
     //determine query time averages
     //List<String> reportIds = GetUniqueReportIds(periodMetrics);
-    double currentQueryTimeAvg = (totalQueryTimeSpent / reportIds.size());
+    double currentQueryTimeAvg = (totalQueryTimeSpent / Integer.max(1, periodQueryMetrics.size()));
```
smailliwcs, line 253: With this change, can now remove `reportIds` from this method's signature.  (And for `calculateValidationTimeAvg` and `calculateEvaluationTimeAvg` below.)

smailliwcs (approved):
See comments re: parameter cleanup, but looks good otherwise.

api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
MontaltoNick: I clean this up in the PR for 1846, actually

PR #687, opened 2024-01-31
Ensure bundle sorting retains all entries

smailliwcs:
From yesterday's discussion on #arch.

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
+++ b/core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
+import java.util.stream.Collectors;
 
 public class FhirBundlerEntrySorter {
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/lantanagroup/link/FhirBundlerEntrySorter.java
```diff
@@ -56,25 +57,24 @@ public static void sort(Bundle bundle) {
     for (String patientId : fhirBundleProcessor.getPatientIds()) {
       logger.debug("Adding patient resources: {}", patientId);
       List<Bundle.BundleEntryComponent> relatedPatientResources = patientResources.get(patientId);
-      Bundle.BundleEntryComponent indMeasureReport = relatedPatientResources.stream()
+      List<Bundle.BundleEntryComponent> indMeasureReports = relatedPatientResources.stream()
```
seanmcilvenna, line 60: In future it would be good to figure out how to identify MeasureReport resources generated by Link in case we get MeasureReport data from the EHR/Facility. We've entertained that idea early on with parkland... using MeasureReports from the facility as a means of indicating inventory data.  Not necessary for now, but something to be aware of.

seanmcilvenna (approved):

PR #688, opened 2024-01-31
LNK-1524: fix validation error

arianamihailescu:


smailliwcs (changes requested):
I don't think we want to map *all* `DataFormatException`s to 400s, since it's possible they'll be produced outside of request binding.  Better strategy might be to wrap the body of `FhirJsonDeserializer.deserialize` in a `try` block and map all exceptions thrown there (not just `DataFormatException`, but any arbitrary `Exception`) to a `ResponseStatusException` with `HttpStatus.BAD_REQUEST`.

seanmcilvenna (approved):

core/src/main/java/com/lantanagroup/link/serialize/FhirJsonDeserializer.java
```diff
@@ -1,23 +1,27 @@
...
   @Override
-  public T deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) throws IOException {
-    String jsonContent = jsonParser.readValueAsTree().toString();
-    return (T) this.jsonParser.parseResource(jsonContent);
+  public T deserialize(JsonParser jsonParser, DeserializationContext deserializationContext) {
+    try {
+      String jsonContent = jsonParser.readValueAsTree().toString();
+      return (T) this.jsonParser.parseResource(jsonContent);
+    } catch (Exception e) {
+      throw new ResponseStatusException(HttpStatus.BAD_REQUEST, e.getMessage());
```
smailliwcs, line 24: Worthwhile to pass `e` as the third argument here so we get the full stack trace from the underlying error in the logs?

smailliwcs (approved):
Looks good, but see comment for potential tweak.

PR #690, opened 2024-02-01
Release 0.3.2

seanmcilvenna:


smailliwcs:
Guessing the intention here is to bring `master` up to date with the latest release?  If so, any objection to just overwriting `master` rather than attempting to merge?

```
# Merge master into rc_0.3.2 with the "ours" merge strategy
# I.e., create a new merge commit with both master and rc_0.3.2 as parents
# But don't actually merge ... just take the content from rc_0.3.2
git checkout rc_0.3.2
git merge -s ours master
# Fast-forward master to the new merge commit
git checkout master
git merge rc_0.3.2
git push origin master
```

Though note that this will move `rc_0.3.2` to the new merge commit as well.  If not desired, could `checkout -b` a temporary branch before the first merge, then `branch -d` it at the end.

seanmcilvenna:
Agreed. That is the intent. We should do this, since it wasn’t done for the last rc_0.3.2 release.


Sean McIlvenna
c: 208-704-6613
North Idaho (Pacific Time)

smailliwcs:
Replaced by #715.

PR #692, opened 2024-02-02
Lnk 1828

arianamihailescu:


core/src/main/java/com/lantanagroup/link/ReportIdHelper.java
```diff
@@ -41,4 +41,8 @@ public static String getPatientMeasureReportId(String masterMeasureReportId, Str
   public static String getPatientDataBundleId(String masterIdentifierValue, String patientId) {
     return combine(masterIdentifierValue, hash(patientId));
   }
+
+  public static String getPatientBundleFileName(String masterIdentifierValue, String patientId) {
+    return combine("patient-bundle-for-cqf", masterIdentifierValue, patientId);
```
seanmcilvenna, line 46: Suggest putting `-` in between the parts...
i.e. `return combine("patient-bundle-", masterIdentifierValue, "-", patientId);`

seanmcilvenna (approved):

core/src/main/java/com/lantanagroup/link/ReportIdHelper.java
arianamihailescu: Added

arianamihailescu: The combine was already adding  - now there is -- between components

PR #698, opened 2024-02-15
LNK-2065: Using NPM packages to load resources for validation

seanmcilvenna:


core/src/main/java/com/lantanagroup/link/validation/Validator.java
```diff
@@ -93,32 +106,84 @@ private static OperationOutcome.IssueType getIssueCode(String messageId) {
...
+   */
+  private void loadTerminology() {
+    try {
+      PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
+      org.springframework.core.io.Resource[] resources = resolver.getResources("terminology/**");
+      this.loadClassResources(resources);
+    } catch (IOException e) {
+      logger.error("Error loading class resources for validation: {}", e.getMessage());
+    }
+
```
smailliwcs, line 186: Leaving this for now, but as noted in the [commit comments](https://github.com/lantanagroup/link/commit/5b0718e8d6e53e7a58552e3ed1e7c37e2d81dd60#r138677329), the only remaining terminology resource appears to be in R4 core (and thus *shouldn't* need to be explicitly loaded into the validation support).

https://simplifier.net/packages/hl7.fhir.r4.core/4.0.1/files/82054

smailliwcs (approved):

PR #701, opened 2024-02-22
LNK-1523: Update measure evaluation to evaluate directly within the API

smailliwcs:


api/src/main/java/com/lantanagroup/link/api/MeasureService.java
```diff
@@ -0,0 +1,79 @@
...
+import org.opencds.cqf.fhir.cql.engine.retrieve.RetrieveSettings;
+import org.opencds.cqf.fhir.cql.engine.terminology.TerminologySettings;
+import org.opencds.cqf.fhir.cr.measure.MeasureEvaluationOptions;
+import org.opencds.cqf.fhir.cr.measure.r4.R4MeasureService;
+import org.opencds.cqf.fhir.utility.monad.Eithers;
+import org.opencds.cqf.fhir.utility.repository.InMemoryFhirRepository;
+
+public class MeasureService {
+  private final MeasureDef measureDef;
+  private final R4MeasureService measureService;
```
seanmcilvenna, line 22: Maybe want to change this variable name, since the class it resides in _is also_ called `MeasureService`. Perhaps something like `hapiMeasureService`. Or maybe even `r4MeasureService`. Not really happy with any of those options, though, frankly.

api/src/main/java/com/lantanagroup/link/api/MeasureService.java
```diff
@@ -0,0 +1,79 @@
...
+  public void preCompile() {
+    evaluate(null, null, null, null);
+  }
+
+  public MeasureReport evaluate(String periodStart, String periodEnd, String subject, Bundle additionalData) {
+    return measureService.evaluate(
+            Eithers.forRight3(measureDef.getMeasure()),
+            periodStart,
+            periodEnd,
+            null,
```
seanmcilvenna, line 68: I'm sure I can find the answer here... But curious what these other parameters are and if they're something we might want to populate later. Just taking note, for now. Nothing to change here.

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -58,9 +58,8 @@ public ReportGenerator(SharedService sharedService, TenantService tenantService,
    * This method accepts a list of patients and generates an individual measure report for each patient.
    */
   public void generate(QueryPhase queryPhase) throws ExecutionException, InterruptedException {
-    if (this.config.getEvaluationService() == null) {
-      throw new IllegalStateException("api.evaluation-service has not been configured");
-    }
+    MeasureService measureService = new MeasureService(measureContext.getReportDefBundle(), config.getTerminologyService());
```
seanmcilvenna, line 61: In future, consider making a wrapping `MeasuresService` (plural) or something like that, that is a SpringBoot singleton `@service`, so that we can don't have to re-pre-compile every time we generate. Not sure how much time `preCompile()` takes, though... So maybe not worth it. But, caching might further improve performance. Then invalidate the cache when measure refresh is requested in the API.

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -58,9 +58,8 @@ public ReportGenerator(SharedService sharedService, TenantService tenantService,
...
+    measureService.preCompile();
     logger.info("Patient list is : " + measureContext.getPatientsOfInterest(queryPhase).size());
     ForkJoinPool forkJoinPool = config.getMeasureEvaluationThreads() != null
```
seanmcilvenna, line 64: @edward-miller-lcg FYI... This PR is big, and touches on files that you'll be looking at editing related to the ForkJoinPool ticket. Consider basing your changes on this PR/branch if it isn't merged before you start working on it.

api/src/main/java/com/lantanagroup/link/api/controller/MeasureDefController.java
```diff
@@ -44,24 +44,6 @@ public void initializeBinder(WebDataBinder binder) {
     binder.setDisallowedFields();
   }
 
-  /**
-   * Executes the measure bundle on the evaluation service (cqf-ruler)
-   */
-  private void executeBundle(Bundle bundle) {
```
seanmcilvenna, line 50: It might be good to have an operation here or in the QAController to allow us to evaluate a `Bundle` of patient data with the configuration of NHSNLink. I suspect we may find value in that, to test issues. We _could_ keep the CQF-Ruler deployments around, but that could potentially give different results because we're using it differently, or different resources are available to it, etc. Maybe something like `POST /api/measureDef/NHSNHypo.../$test?start=XX&end=YY`, which returns a MeasureReport.

seanmcilvenna (approved):
Fantastic! Nothing stood out to me as really problematic. Only made comments about "future considerations".

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
smailliwcs: For hypo, eager compilation saves something like three seconds on the first evaluation.  (It also gives us peace of mind that there won't be threading issues if that first evaluation is actually _multiple simultaneous_ evaluations, which it generally is.)  So absolutely worthwhile to cache across patients (spinning up the `MeasureService` here in `ReportGenerator` rather than `MeasureEvaluator`), but probably not worthwhile to cache across requests just to save a few seconds per `$generate`.

PR #703, opened 2024-02-23
LNK-2037

MontaltoNick:
Add Report Size Metrics


core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
```diff
@@ -32,6 +35,31 @@ public PatientMeasureReport findById(String id) {
...
+            .reduce(StreamUtils::toOnlyElement)
+            .orElse(null);
+  }
+
+  public List<PatientMeasureReportSize> GetMeasureReportSize(String patientId, String reportId, String measureId) {
+    String sql =  "SELECT pmr.patientId, pmr.reportId, pmr.measureID, CAST(SUM(DATALENGTH(pmr.measureReport)) as FLOAT)/1024.0 as sizeKb FROM dbo.patientMeasureReport AS pmr WHERE (ISNULL(:reportId, '') = '' OR reportId = :reportId) AND (ISNULL(:measureId, '') = '' OR measureId = :measureId) AND (ISNULL(:patientId, '') = ''  OR patientId = :patientId) GROUP BY pmr.patientId, pmr.reportId, pmr.measureId";
+
+    if(reportId == null) reportId = "";
+    if(measureId == null) measureId = "";
+    if(patientId == null) patientId = "";
```
smailliwcs, line 53: Duplicative with the `ISNULL` logic in the SQL query itself.  Could remove one or the other.  (Or both and just check for, say, `:reportId IS NULL OR reportId = :reportId` in the query.)

api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
```diff
@@ -401,4 +402,78 @@ private List<MetricData> getPeriodMetricData(String category, String task, List<
...
+          @RequestParam(name = "patientId", required = false) String patientId,
+          @RequestParam(name = "reportId", required = false) String reportId,
+          @RequestParam(name = "measureId", required = false) String measureId) {
+
+    var tenantService = TenantService.create(sharedService, tenantId);
+
+    var reportSizes = tenantService.getPatientMeasureReportSize(patientId, reportId, measureId);
+
+    var summary = new PatientMeasureReportSizeSummary();
+    Double sum = 0.0;
```
smailliwcs, line 417: Do we need a `Double` object here (as opposed to a `double` primitive)?  Best to use primitives where possible to avoid needless conversion to/from the object wrapper.

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
```diff
@@ -32,6 +35,31 @@ public PatientMeasureReport findById(String id) {
             .orElse(null);
   }
 
+  public PatientMeasureReportSize GetMeasureReportSizeById(String id) {
```
smailliwcs, line 38: Where is this used?  (I see that the one-arg variant of `TenantService.getPatientMeasureReportSize` calls it, but then where is *that* used?)

core/src/main/java/com/lantanagroup/link/db/mappers/BaseMapper.java
```diff
@@ -57,7 +57,9 @@ public Row(ResultSet resultSet) {
     public String getString(String columnName) throws SQLException {
       return resultSet.getNString(columnName);
     }
-
+    public Double getDouble(String columnName) throws SQLException {
```
smailliwcs, line 60: Same comment re: object vs. primitive.

core/src/main/java/com/lantanagroup/link/model/PatientMeasureReportSizeSummary.java
```diff
@@ -0,0 +1,26 @@
...
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.HashMap;
+import java.util.List;
+
+@Getter
+@Setter
+public class PatientMeasureReportSizeSummary{
```
smailliwcs, line 12: Same comment re: object vs. primitive (for the two `Double` fields).

api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
```diff
@@ -401,4 +402,78 @@ private List<MetricData> getPeriodMetricData(String category, String task, List<
...
+
+      sum += r.getSizeKb();
+
+      if(measureCountMap.containsKey(mId))
+        measureCountMap.put(mId, measureCountMap.get(mId) + 1);
+      else {
+        measureCountMap.put(mId, 1);
+        averageReportSizeByMeasureId.put(mId, 0.0);
+      }
+      averageReportSizeByMeasureId.put(mId, averageReportSizeByMeasureId.get(mId) + r.getSizeKb());
```
smailliwcs, line 440: Not necessary to change, but `Map` offers some more expressive ways to achieve this kind of thing:

```java
measureCountMap.merge(mId, 1, (count, value) -> count + value);
averageReportSizeByMeasureId.merge(mId, r.getSizeKb(), (size, value) -> size + value);
```

See also `compute`, `computeIfAbsent`, `computeIfPresent`, `putIfAbsent`, etc.

api/src/main/java/com/lantanagroup/link/api/controller/MetricController.java
```diff
@@ -401,4 +402,78 @@ private List<MetricData> getPeriodMetricData(String category, String task, List<
...
+    }
+
+    summary.setTotalSize(sum);
+    summary.setReports(reportSizes);
+    summary.setReportCount(reportSizes.size());
+    summary.setAverageReportSize(sum/reportSizes.size());
+
+    for(var kv : measureCountMap.entrySet()) {
+      averageReportSizeByMeasureId.put(kv.getKey(), averageReportSizeByMeasureId.get(kv.getKey())/kv.getValue());
+    }
```
smailliwcs, line 467: Similar to the previous comment ... might look into `Map.replaceAll`?

smailliwcs (changes requested):
Feel free to ignore any of the inline comments, but the ticket mentions reporting on the size of resources in `dbo.patientData`, which doesn't appear to be addressed here.

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
MontaltoNick: I thought so, too, but it was throwing an exception when I tried to fill the parameter with the null string, so I added this code to get past it. If I then change the query to only check if its null or has a value of the reportId, the where clause will always faill, because it isn't NULL and the id column will never be "".

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
MontaltoNick: It isn't used. I wrote up the back end in case we might want to have that functionality wired up to the front end. I could wire it up to an endpoint or delete it. I wasn't sure if this use case would be useful.

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
MontaltoNick: I will remove the ISNULL component from the query. Should work still and make the sql slightly shorter.

smailliwcs: Ah, yeah, `Map.of` doesn't allow null keys/values, if that's where you were hitting the exception.  Kinda annoying, IMO.

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
smailliwcs: Not a problem to leave it as is, IMO.  (Or wire it up or delete it.)  Just wanted to verify it wasn't supposed to be called somewhere.

MontaltoNick: I implemented it and created the postman endpoint configuration for it.

core/src/main/java/com/lantanagroup/link/db/repositories/PatientMeasureReportRepository.java
MontaltoNick: Yes, very annoying.

smailliwcs (approved):

PR #706, opened 2024-02-27
Lnk 1851 use common pool reports

edward-miller-lcg:


api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -62,9 +62,7 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
       throw new IllegalStateException("api.evaluation-service has not been configured");
     }
     logger.info("Patient list is : " + measureContext.getPatientsOfInterest(queryPhase).size());
-    ForkJoinPool forkJoinPool = config.getMeasureEvaluationThreads() != null
-            ? new ForkJoinPool(config.getMeasureEvaluationThreads())
-            : ForkJoinPool.commonPool();
```
smailliwcs, line 67: Should be able to remove `ApiConfig.measureEvaluationThreads`.

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
```diff
@@ -62,9 +62,7 @@ public void generate(QueryPhase queryPhase) throws ExecutionException, Interrupt
...
+    ForkJoinPool forkJoinPool = ForkJoinPool.commonPool();
```
smailliwcs, line 65: No longer necessary to `shutdown` the pool (see `finally` below).

query/src/main/java/com/lantanagroup/link/query/uscore/PatientScoop.java
```diff
@@ -105,8 +106,9 @@ public void loadInitialPatientData(ReportCriteria criteria, ReportContext contex
     // first get the patients and store them in the patientMap
     Map<String, Patient> patientMap = new ConcurrentHashMap<>();
     int threshold = this.tenantService.getConfig().getFhirQuery().getParallelPatients();
```
smailliwcs, line 108: Should be able to remove `FhirQuery.parallelPatients` (and this line).

query/src/main/java/com/lantanagroup/link/query/uscore/PatientScoop.java
```diff
@@ -105,8 +106,9 @@ public void loadInitialPatientData(ReportCriteria criteria, ReportContext contex
...
-    ForkJoinPool patientDataFork = new ForkJoinPool(threshold);
-    ForkJoinPool patientFork = new ForkJoinPool(threshold);
+    ForkJoinPool patientDataFork = ForkJoinPool.commonPool();
+    ForkJoinPool patientFork = ForkJoinPool.commonPool();
```
smailliwcs, line 110: As above ... no longer necessary to `shutdown` these pools.

smailliwcs (changes requested):

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
edward-miller-lcg: this was removed. this is the old code for the compare.

query/src/main/java/com/lantanagroup/link/query/uscore/PatientScoop.java
edward-miller-lcg: done

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
smailliwcs: Sorry, I was referring to the underlying config property in the `core` module (`com.lantanagroup.link.config.api.ApiConfig.measureEvaluationThreads`).  If there's nothing else using that (and there *shouldn't* be, I don't think), then we should remove it, since that bit of configurability doesn't exist anymore.

query/src/main/java/com/lantanagroup/link/query/uscore/PatientScoop.java
smailliwcs: As above, but this one is in tenant config (`com.lantanagroup.link.db.model.tenant.FhirQuery.parallelPatients`).  Ideally, we'd remove that field, since it's no longer used.  (As long as that doesn't break de/serialization.)

api/src/main/java/com/lantanagroup/link/api/ReportGenerator.java
edward-miller-lcg: got it. Removed that property.

smailliwcs (approved):

PR #708, opened 2024-03-01
Misc improvements to Device generation

seanmcilvenna:
* Keeping track of ImplementationGuide resources found in NpmPackages within the `Validator` separately from the `PrePopulatedValidationSupport` since `PrePopulatedValidationSupport` doesn't do anything with ImplementationGuide
* Loading Device.notes with list of implementation guides that are loaded at the time of validation
* Update Device.notes in the report when a report is re-validated
* Changing interface for IPatientIdProvider to not return the patient list, since it just loads it into the report context anyways and the return value isn't used.
* Change `Device` generated to use `note` for events and concept maps. Much more readable. Conferred with Karen and Dave, who agreed.


core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -270,54 +274,68 @@ public static Device getDevice(ApiConfig apiConfig) {
...
+  }
+
+  private static void addEventNotesToDevice(Device device, String eventCategory, List<String> events) {
     if (events == null) {
       return;
     }
 
     for (String event : events) {
-      Device.DevicePropertyComponent property = device.addProperty();
-      property.getType().addCoding().setCode("event").setSystem(Constants.LinkDevicePropertiesCodeSystem);
```
smailliwcs, line 317: If our strategy is changing from `property` to `note`, should we remove `Constants.LinkDevicePropertiesCodeSystem`?  (And does that change need to flow up to the IG?)

api/src/main/java/com/lantanagroup/link/api/controller/ValidationController.java
```diff
@@ -324,6 +328,14 @@ public OperationOutcome validate(@PathVariable String tenantId, @PathVariable St
...
     }
 
+    // Update the report's device with the current list of implementation guides
+    Device device = report.getDeviceInfo();
+    if (device == null) {
+      device = FhirHelper.getDevice(this.apiConfig);
+    }
+    FhirHelper.setImplementationGuideNotes(device, this.validator);
+    tenantService.saveReport(report);
+
```
smailliwcs, line 338: Doesn't IG reporting really belong in the validation output, rather than in the report output?

  - We don't use the IGs during report generation, only during validation.
  - We could technically validate the same report against different sets of IGs.  (Would require the IGs being updated between those validations ... an edge case, for sure.)

Prior to LNK-1336, we'd been adding `ImplementationGuide` resources to the validation output `Bundle`.  Would require a different strategy now, since the output is an `OperationOutcome`.

core/src/main/java/com/lantanagroup/link/FhirHelper.java
```diff
@@ -270,54 +274,68 @@ public static Device getDevice(ApiConfig apiConfig) {
...
     }
 
     List<com.lantanagroup.link.db.model.ConceptMap> conceptMaps = tenantService.getAllConceptMaps();
 
     if (!conceptMaps.isEmpty()) {
-      Device.DevicePropertyComponent property = device.addProperty();
-      property.setType(new CodeableConcept().addCoding(new Coding().setCode("concept-map").setSystem(Constants.LinkDevicePropertiesCodeSystem)));
+      for (ConceptMap conceptMap : conceptMaps) {
+        String title = conceptMap.getConceptMap().getTitle();
+        String version = StringUtils.isNotEmpty(conceptMap.getConceptMap().getVersion()) ? " (" + conceptMap.getConceptMap().getVersion() + ")" : "";
```
smailliwcs, line 305: Definitely an improvement to pull in the version, but I doubt we're being diligent enough about setting its value for it to be useful.  Could consider hashing the content instead?  Though we might have to think about pre-hash formatting for even _that_ to be useful.

core/src/main/java/com/lantanagroup/link/validation/Validator.java
```diff
@@ -150,6 +149,19 @@ public void init() {
...
   }
 
+  private static ImplementationGuide getImplementationGuide(ImplementationGuide baseResource) {
+    ImplementationGuide ig = baseResource;
+    ImplementationGuide igToAdd = new ImplementationGuide();
+    igToAdd.setUrl(ig.getUrl());
+    igToAdd.setVersion(ig.getVersion());
+    igToAdd.setName(ig.getName());
+    igToAdd.setTitle(ig.getTitle());
+    igToAdd.setStatus(Enumerations.PublicationStatus.UNKNOWN);
```
smailliwcs, line 159: Curious about the special case here.  Why is `status` the only element that doesn't flow through from the original IG resource?

core/src/main/java/com/lantanagroup/link/validation/Validator.java
```diff
@@ -176,8 +188,19 @@ private void loadPackages() {
         npmPackage.listResources(allowsResourceTypes).forEach(resource -> {
           try {
             String resourceJson = new String(npmPackage.getFolders().get("package").getContent().get(resource), StandardCharsets.UTF_8);
-            IBaseResource baseResource = this.jsonParser.parseResource(resourceJson);
-            this.prePopulatedValidationSupport.addResource(baseResource);
+            Resource baseResource = (Resource) this.jsonParser.parseResource(resourceJson);
```
smailliwcs, line 191: Not necessary to change, but possible more performant to parse these resources in a streaming fashion (rather than sucking them into a `String` and parsing that):

```java
Resource baseResource;
try (InputStream stream = npmPackage.load(resource)) {
    baseResource = (Resource) this.jsonParser.parseResource(stream);
}
```

This also avoids the need to specify how to traverse through the package structure.

smailliwcs (approved):

core/src/main/java/com/lantanagroup/link/FhirHelper.java
seanmcilvenna: Yah I had a discussion with @dadmbc and @karenmbourque about this a bit... The version of the nhsn-measures IG is simply "1.0.0-cibuild" right now. Measure dev team would need to change their processes a bit to increment the version of the IG routinely for this information to be useful. I'd _like_ to expect that we can rely on it in the future. Maybe not yet, but soon?

core/src/main/java/com/lantanagroup/link/validation/Validator.java
seanmcilvenna: No reason :) I think that's how it was in the original iteration of outputing the IGs. I'm going to update the PR to bring status forward.

core/src/main/java/com/lantanagroup/link/FhirHelper.java
smailliwcs: Wouldn't this version be manually assigned by whoever developed/posted the concept map?  (Which would ideally be a datestamp of its last meaningful update, but inevitably is copy/pasted from some _other_ concept map and then never bumped.)

core/src/main/java/com/lantanagroup/link/validation/Validator.java
seanmcilvenna: I like it. Incorporated!

PR #717, opened 2024-03-08
New endpoint to get both categories and results together

seanmcilvenna:
... together in a single request/response/file where an .XSLT *could* be used to convert it into an .html file that can be shared with stakeholders (something similar is already occuring, but with "valview" python scripts that I'd like to deprecate the use of.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -67,6 +67,11 @@
             <artifactId>hapi-fhir-client</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-xml</artifactId>
+            <version>2.16.1</version>
+        </dependency>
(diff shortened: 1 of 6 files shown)


core/pom.xml
```diff
@@ -66,6 +66,11 @@
             <groupId>ca.uhn.hapi.fhir</groupId>
             <artifactId>hapi-fhir-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-xml</artifactId>
+            <version>2.16.1</version>
```
smailliwcs, line 72: Would prefer that we don't specify an explicit version here, instead allowing the correct/compatible/consistent version to flow in from the parent POM's `dependencyManagement`.

smailliwcs (approved):

PR #720, opened 2024-03-12
Creating app config that can be altered in each deployed environment

seanmcilvenna:
For authentication and the base url of the API
To customize per each environment, need to overwrite (or mount) a custom `app.config.json` file to the deployed assets directory within the pod/container.
Default configuration points to DEV for the API.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -7,3 +7,5 @@
 # Editor-based HTTP Client requests
 /httpRequests/
 /kubernetes-settings.xml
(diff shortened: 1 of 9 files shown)


seanmcilvenna:
FYI, I already updated the DEV and TEST kubernetes deployments to have these config files mounted to the assets folder, so that if/when this was merged, they would be ready to go with the env-specific configuration

smailliwcs (approved):

PR #721, opened 2024-03-14
LNK-2186 Validation category and results details HTML output

seanmcilvenna:
* Moving `ValidationCategorizer` to the validation package.
* Adding unit test to ensure unique category ids are specified.
* Updates to categories to de-duplicate them, fixing the naming/guidance on at least one of them.
* Add text/html output for the validation category results endpoint that produces an HTML file that can be downloaded and opened in a browser, showing the categories and validation result details.
* Moving some logic for categorizing from the controller to `ValidationCategorizer`.
* Adding validation-results.html as output in the folder-based submission so that they can be automatically generated and included in the submission for review.


core/src/main/java/com/lantanagroup/link/sender/FileSystemSender.java
```diff
@@ -217,6 +217,13 @@ private void saveToFolder(Bundle bundle, String path) throws Exception {
     if (otherResourcesBundle.hasEntry()) {
       this.saveToFile(otherResourcesBundle, Paths.get(path, "other-resources.json").toString());
     }
+
+    // Save validation results as HTML
+    logger.debug("Saving validation results as HTML");
+    String html = new ValidationCategorizer().getValidationCategoriesAndResultsHtml(tenantService, report);
+    if (StringUtils.isNotEmpty(html)) {
+      this.saveToFile(html.getBytes(StandardCharsets.UTF_8), Paths.get(path, "validation-report.html").toString());
+    }
```
smailliwcs, line 226: Guessing this will become apparent when you attempt to merge, but as part of LNK-1871, there was a change to automatically save validation results (the `OperationOutcome`) in the same place as the submission bundle even when using "unified bundling" (i.e., regardless of whether we hit `saveToFolder` or `saveToFile`).  Would probably be good for this validation report to be saved in `saveToFile` as well.

smailliwcs (approved):

core/src/main/java/com/lantanagroup/link/sender/FileSystemSender.java
seanmcilvenna: Updated so that it stores the .html next to the single-file bundle submission just like the `validation-results.json` file, but as `validation-results.html`

PR #724, opened 2024-03-14
Tweaks to remove the assumption of millisecond precision in reporting period start/end

smailliwcs:


seanmcilvenna (approved):
A significantly simplified `parseFhirDate()` implementation 🤣 👍

PR #731, opened 2024-03-16
Updates to dependencies based on dependabot PRs

seanmcilvenna:


smailliwcs:
Should this be merged into `development` rather than `master`?  And more generally, shouldn't Dependabot be *targeting* `development`?  Especially since we want `master` to track the state of the latest release.

seanmcilvenna:
Good ideas. Not sure how to get dependabot to target development, but I'll re-create the PR for development

PR #733, opened 2024-03-20
LNK-2177: MVP: Update NHSNLink code to prevent regeneration of a report for a site if one is already in progress

smailliwcs:
Update `ReportController.generateResponse` to require acquisition of a tenant-specific lock.


seanmcilvenna (approved):
What a beautifully simple and elegant solution!

smailliwcs:
Didn't even need to reindent! 😂

PR #736, opened 2024-03-27
Remove `originalResource` from `dataTrace`

smailliwcs:


core/src/main/resources/tenant-db.sql
```diff
@@ -125,6 +125,11 @@ IF NOT EXISTS (SELECT *
 
 GO
 
+ALTER TABLE dbo.dataTrace
```
seanmcilvenna, line 128: @sdmcgeown in the 0.4.1 patch, this `ALTER TABLE` line is what we will need to send to the DB team to have them apply the schema changes _when_ we're ready to upgrade from 0.4.0 to 0.4.1 in beta.

seanmcilvenna (approved):

core/src/main/resources/tenant-db.sql
sdmcgeown: @seanmcilvenna Understood

PR #740, opened 2024-04-02
LNK-1547: Adding in basic versioning to metric entries in SQL databas…

ajagann1:
…e so it's easier to determine which metrics pertain to which version of the generated report

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
+++ b/api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
@@ -321,6 +321,7 @@ public class ReportController extends BaseController {
    */
   private Report generateResponse(TenantService tenantService, LinkCredentials user, HttpServletRequest request, String packageId, List<String> measureIds, String periodStart, String periodEnd, boolean regenerate, boolean validate, boolean skipQuery, List<String>  debugPatients) throws Exception {
     Report report = null;
+    String currentVersion = "";
     Lock tenantLock = tenantLocks.computeIfAbsent(tenantService.getConfig().getId(), id -> new ReentrantLock());
     if (!tenantLock.tryLock()) {
       throw new ResponseStatusException(HttpStatus.CONFLICT, "Report in progress for tenant");
@@ -442,9 +443,12 @@ public class ReportController extends BaseController {
       this.sharedService.audit(user, request, tenantService, AuditTypes.Generate, String.format("Generated report %s", report.getId()));
       logger.info("Done generating report {}, continuing to bundle and validate...", report.getId());
 
+      //New report generation means increment the version
+      currentVersion = String.valueOf(Integer.parseInt(this.sharedService.getCurrentMetricVersion(report.getId())) + 1);
+
       if (validate) {
         try {
-          this.validationService.validate(stopwatchManager, tenantService, report);
+          this.validationService.validate(stopwatchManager, tenantService, report, currentVersion);
           logger.info("Done validating report");
         } catch (Exception ex) {
           logger.error("Error validating report {}", report.getId(), ex);
@@ -456,7 +460,8 @@ public class ReportController extends BaseController {
       tenantLock.unlock();
     }
 
-    this.stopwatchManager.storeMetrics(tenantService.getConfig().getId(), report.getId());
+    this.stopwatchManager.storeMetrics(tenantService.getConfig().getId(), report.getId(),
+            currentVersion);
     logger.info("Statistics for report {} are:\n{}", report.getId(), this.stopwatchManager.getStatistics());
     this.stopwatchManager.reset();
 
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/lantanagroup/link/db/SharedService.java
```diff
@@ -447,10 +447,11 @@ public void saveMetrics(List<Metrics> metrics){
     metrics.forEach(metric -> {
       try (Connection conn = this.getSQLConnection()) {
         assert conn != null;
-        SQLCSHelper cs = new SQLCSHelper(conn, "{ CALL saveMetrics (?, ?, ?, ?, ?, ?, ?) }");
+        SQLCSHelper cs = new SQLCSHelper(conn, "{ CALL saveMetrics (?, ?, ?, ?, ?, ?, ?, ?) }");
         cs.setNString("id", metric.getId().toString());
         cs.setNString("tenantId", metric.getTenantId());
         cs.setNString("reportId", metric.getReportId());
+        cs.setNString("versionNum", metric.getVersionNum());
```
seanmcilvenna, line 454: Rename to just `version`

core/src/main/java/com/lantanagroup/link/time/StopwatchManager.java
```diff
@@ -77,6 +77,7 @@ public synchronized void storeMetrics(String tenantId, String reportId) {
       Metrics metric = new Metrics();
       metric.setTenantId(tenantId);
       metric.setReportId(reportId);
+      metric.setVersionNum(version);
```
seanmcilvenna, line 80: Rename to `version`

core/src/main/java/com/lantanagroup/link/validation/ValidationService.java
```diff
@@ -86,6 +86,7 @@ public OperationOutcome validate(StopwatchManager stopwatchManager, TenantServic
     String category = Constants.VALIDATION_ISSUE_CATEGORY;
     metric.setTenantId(tenantService.getConfig().getId());
     metric.setReportId(report.getId());
+    metric.setVersionNum(version);
```
seanmcilvenna, line 89: Rename

core/src/main/resources/shared-db.sql
```diff
@@ -89,6 +89,7 @@ IF OBJECT_ID(N'dbo.metrics', N'U') IS NULL
             id           UNIQUEIDENTIFIER NOT NULL PRIMARY KEY DEFAULT NEWID(),
             tenantId     NVARCHAR(128)    NOT NULL,
             reportId     NVARCHAR(128)    NOT NULL,
+            versionNum   NVARCHAR(128),
```
seanmcilvenna, line 92: This .sql file is maintained over time... We can't got back and alter the original "CREATE TABLE" statement, because this has already been executed on other environments. Instead, at the end of this .sql file, you should check if the column exists or not, and then add it to the table if it does not. An example of this appears in `tenant-db.sql` like so:

```
IF (SELECT DATA_TYPE
    FROM INFORMATION_SCHEMA.COLUMNS
    WHERE TABLE_SCHEMA = 'dbo'
      AND TABLE_NAME = 'bulkStatusResult'
      AND COLUMN_NAME = 'statusId') = 'nvarchar'
    BEGIN
        ALTER TABLE dbo.bulkStatusResult
            ALTER COLUMN statusId uniqueidentifier NULL;
        ALTER TABLE dbo.bulkStatusResult
            ADD FOREIGN KEY (statusId) REFERENCES dbo.bulkStatus (id);
    END
GO
```

core/src/main/resources/shared-db.sql
```diff
@@ -280,22 +282,6 @@ BEGIN
 END
 GO
 
--- LNK-1359
```
seanmcilvenna, line 283: Don't think you intended to remove the `saveMetrics` SPROC here...
If anything, this SPROC needs to be updated to take into account `@version` parameter, and including `version` column in the `INSERT` statement

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -442,9 +443,12 @@ private Report generateResponse(TenantService tenantService, LinkCredentials use
       this.sharedService.audit(user, request, tenantService, AuditTypes.Generate, String.format("Generated report %s", report.getId()));
       logger.info("Done generating report {}, continuing to bundle and validate...", report.getId());
 
+      //New report generation means increment the version
+      currentVersion = String.valueOf(Integer.parseInt(this.sharedService.getCurrentMetricVersion(report.getId())) + 1);
```
seanmcilvenna, line 447: I believe we already have a version for the report set in `Report.version`. Instead of creating a new `currentVersion` I would simply re-use `report.getVersion()`. However, that version number is a string, not a number. i.e. "1.0", "1.1", "2.0", "2.1"

seanmcilvenna (approved):

PR #745, opened 2024-04-17
LNK-2334

MontaltoNick:
Remove getFile and GetFileName function calls, use more specific io so that we don't have to investigate file extensions.

### DIFF
--- a/core/src/main/java/com/lantanagroup/link/validation/Validator.java
+++ b/core/src/main/java/com/lantanagroup/link/validation/Validator.java
@@ -229,15 +229,15 @@ public class Validator {
       PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
       org.springframework.core.io.Resource[] xmlResources = resolver.getResources("terminology/*.xml");
       org.springframework.core.io.Resource[] jsonResources = resolver.getResources("terminology/*.json");
-      this.loadresources(xmlResources, this.xmlParser);
-      this.loadresources(jsonResources, this.jsonParser);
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/lantanagroup/link/validation/Validator.java
```diff
@@ -232,45 +227,62 @@ private void loadPackages() {
   private void loadTerminology() {
     try {
       PathMatchingResourcePatternResolver resolver = new PathMatchingResourcePatternResolver();
-      org.springframework.core.io.Resource[] resources = resolver.getResources("terminology/**");
-      this.loadClassResources(resources);
+      org.springframework.core.io.Resource[] xmlResources = resolver.getResources("terminology/*.xml");
+      org.springframework.core.io.Resource[] jsonResources = resolver.getResources("terminology/*.json");
+      this.loadXmlResource(xmlResources);
+      this.loadJsonResources(jsonResources);
```
smailliwcs, line 233: Looks like the implementations of these two methods are mostly the same.  Recommend collapsing to a single `loadResources` method with an additional parameter: an `IParser` instance that we pass in as either `this.xmlParser` or `this.jsonParser` (depending on whether we're loading `xmlResources` or `jsonResources`).

Also note that `loadXmlResource` is missing the bit where we add the parsed resources to the `prePopulatedValidationSupport`.  So recommend removing `loadXmlResource` and tweaking `loadJsonResources` as described above.

smailliwcs (changes requested):

smailliwcs (approved):

PR #747, opened 2024-05-07
LNK-2298: Get report insights (metrics and errors) by invoking API endpoint

smailliwcs:


api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
```diff
@@ -666,6 +666,25 @@ public MeasureReport getPatientMeasureReport(@PathVariable String patientMeasure
...
+
+    String styles;
+    try {
+      styles = IOUtils.resourceToString("/insights.css", StandardCharsets.UTF_8);
+    } catch (IOException e) {
+      styles = "";
+    }
+    String content = this.sharedService.getReportInsights(tenantId, reportId, version)
+            + tenantService.getReportInsights(tenantId, reportId, version);
+    return String.format("<style type=\"text/css\">%s</style><div class=\"container\">%s</div>", styles, content);
```
seanmcilvenna, line 685: Does this need an `<html><body></body></html>` wrapping it? Guessing the browser figures out how to render it ok... So maybe no big deal.

seanmcilvenna (approved):

api/src/main/java/com/lantanagroup/link/api/controller/ReportController.java
smailliwcs: Not needed, but lazy not to include it.  Tweaked to add `html`, `head`, and `body`.

PR #757, opened 2024-06-14
Fix folder-based submission bug

smailliwcs:
`Files.move` is only appropriate for empty directories.  Use `FileUtils.moveDirectory` (Apache Commons) instead.


amphillipsLGC (approved):
Looks good

PR #769, opened 2024-08-27
LNK-2800: Adding measureScore equivalent to IP count to individual co…

ajagann1:
…hort measure reports

## Summary by CodeRabbit

- **New Features**
	- Enhanced measure report generation logic to better handle scoring types, particularly for cohort measures.
	- Introduced clearer data absence indications for groups with null measure scores.

- **Documentation**
	- Added comments to clarify assumptions about hybrid scoring types and reinforce the intent behind the logic adjustments.

### DIFF
--- a/api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
+++ b/api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
@@ -111,19 +111,18 @@ public class MeasureEvaluator {
 
       //This logic assumes that there's no hybrid scoring types for a single measure (not even sure it's allowed to do so)
       // group.measureScore is required for DEQM profile validation of non-cohort measures
-      if(measure.getScoring().hasCoding()){
-        if(measure.getScoring().getCoding().stream().noneMatch(c -> c.hasCode() && c.getCode().equals("cohort"))) {
-          measureReport.getGroup().stream()
-                  .filter(g -> g.getMeasureScore() == null || g.getMeasureScore().getValue() == null)
-                  .forEach(g -> g.getMeasureScore()
-                          .addExtension(Constants.DataAbsentReasonExtensionUrl, new CodeType(Constants.DataAbsentReasonUnknownCode)));
-        }
-        //Setting measureScore to count retrieved from IP
-        else{
-          measureReport.getGroup().forEach(g ->
-                  //Cohort measure should only have one population (IP)
-                  g.setMeasureScore(new Quantity(g.getPopulation().get(0).getCount())));
-        }
+      if (measure.getScoring().hasCoding()) {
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -109,15 +109,22 @@
...
+      //This logic assumes that there's no hybrid scoring types for a single measure (not even sure it's allowed to do so)
       // group.measureScore is required for DEQM profile validation of non-cohort measures
-      if(measure.getScoring().hasCoding() &&
-              measure.getScoring().getCoding().stream().noneMatch(c -> c.hasCode() && c.getCode().equals("cohort"))) {
-        measureReport.getGroup().stream()
-                .filter(g -> g.getMeasureScore() == null || g.getMeasureScore().getValue() == null)
-                .forEach(g -> {
-                  g.getMeasureScore().addExtension(Constants.DataAbsentReasonExtensionUrl, new CodeType(Constants.DataAbsentReasonUnknownCode));
-                });
+      if (measure.getScoring().hasCoding()) {
```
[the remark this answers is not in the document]
smailliwcs: Don't think CodeRabbit's suggestion is necessary to implement.  HAPI will auto-create a `CodeableConcept` to return from `getScoring` if necessary.

api/src/main/java/com/lantanagroup/link/api/MeasureEvaluator.java
```diff
@@ -108,16 +108,21 @@ private MeasureReport generateMeasureReport() {
...
+        measureReport.getGroup().forEach(g -> {
+          if (measure.getScoring().getCoding().stream().noneMatch(c -> c.hasCode() && c.getCode().equals("cohort"))) {
+            if (g.getMeasureScore() == null || g.getMeasureScore().getValue() == null) {
+              g.getMeasureScore().addExtension(Constants.DataAbsentReasonExtensionUrl, new CodeType(Constants.DataAbsentReasonUnknownCode));
+            }
+          } else {
+            // Cohort measure should only have one population (IP)
+            g.setMeasureScore(new Quantity(g.getPopulation().get(0).getCount()));
```
smailliwcs, line 121: Maybe a "bad data" situation, but what if the group has no populations?  Do we need to guard against that case?  Maybe supply zero for both the population count and the measure score?

ajagann1: Hm... I'm not sure there would ever be a scenario where you'd intentionally want to have a group created in a Measure resource that doesn't have any populations defined within it.  That said, there's no harm in checking for it so I'll add in the null check.

PR #770, opened 2024-08-27
LNK-2913 - Up-rev to version 0.4.3

sdmcgeown:
@smailliwcs I know it's late in the game, and probably doesn't matter that much, but I removed the _-SNAPSHOT_ from the version in the RC branch since it's not really the right convention for this use case. Let me know if that's ok with you, though.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <artifactId>link</artifactId>
         <groupId>com.lantanagroup.link</groupId>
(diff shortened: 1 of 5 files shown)


smailliwcs (approved):
Agree with removing the SNAPSHOT qualifier.  Ideally, once we "accept" the release and merge/tag it in `master`, we'll also bump the versions in `development` to 0.4.4-SNAPSHOT.

PR #780, opened 2024-11-14
LNK-3073: Up-rev to v0.5.0 in rc_0.5.0 branch

sdmcgeown:


sdmcgeown:
Bypassing rule to include a reviewer since change is trivial

PR #783, opened 2024-11-20
Move cache initialization to `apiInit`

smailliwcs:
`ApiApplication.main` isn't invoked when running in a servlet container, which impacts our Docker images.  Move cache initialization to `apiInit` to ensure it occurs regardless of deployment method.


ajagann1 (approved):
👍

PR #794, opened 2025-02-27
IMPS-698: Resolve the FHIR IDs so they do not persist as full URLs

smailliwcs:
Rather than relativizing with respect to the configured base URL (which may differ from the *reported* base URL, causing relativization to fail), simply delegate ID parsing to HAPI.

## Summary by CodeRabbit

- **Refactor**
  - Streamlined the process for converting patient identifiers by eliminating redundant parameters.
  - Optimized the extraction of identifier values directly from resource entries.
  - Enhanced logging to alert when expected identifier details are missing.


bstewartlg (approved):

kissalk (approved):
Reviewed

